PR #53, opened 2020-05-18
Prompt for a consumer group name when creating a kafka pipeline

AntonZelenin:


agent/src/agent/pipeline/config_handlers/base.py
```diff
@@ -60,11 +60,26 @@ def override_base_config(self, client_config, new_uuid=None, new_pipeline_title=
...
-        if 'library' in self.client_config['source']['config']:
-            self.config['stages'][0]['library'] = self.client_config['source']['config']['library']
+        if 'library' in self.pipeline.source.config:
+            self.config['stages'][0]['library'] = self.pipeline.source.config['library']
+        self.__override_source_using_pipeline()
+        self.__write_source_config_to_stage()
+
+    def __override_source_using_pipeline(self):
+        for k, v in self.pipeline.override_source.items():
+            if not self.__is_in_base_source_config(k):
```
dariakharlan, line 70: Why do you need to check it if the same thing is going to be checked for all config on line 81?

agent/src/agent/pipeline/pipeline.py
```diff
@@ -22,6 +23,7 @@ def __init__(self, pipeline_id: str,
         self.source = source_obj
         self.destination = destination
         self.old_config = None
+        self.override_source = config.pop(self.OVERRIDE_SOURCE) if self.OVERRIDE_SOURCE in config else {}
```
dariakharlan, line 26: `config.pop(self.OVERRIDE_SOURCE, {})`
https://docs.python.org/3/library/stdtypes.html#dict.pop

agent/src/agent/pipeline/pipeline.py
```diff
@@ -30,9 +32,9 @@ def file_path(self) -> str:
     def to_dict(self):
         return {
             **self.config,
+            self.OVERRIDE_SOURCE: self.override_source,
             'pipeline_id': self.id,
-            'source': self.source.to_dict() if self.source else None,
-            'destination': self.destination.to_dict()
+            'source_name': self.source.name,
```
dariakharlan, line 37: Backward compatibility will be broken if you add source_name instead of ['source']['name']

PR #59, opened 2020-05-25
Created a script to run agent

AntonZelenin:


scripts/agent.sh
```diff
@@ -0,0 +1,23 @@
+if [[ $1 == 'install' ]]; then
+  docker-compose pull && docker-compose up -d
+elif [[ $1 == 'run' || $1 == 'rerun' ]]; then
+    docker-compose up -d
+elif [[ $1 == 'attach' ]]; then
```
dariakharlan, line 5: let's maybe do `run` command with `docker-compose up -d && docker attach anodot-agent` and
`upgrade $version` with setting new tag in a docker file and doing `docker-compose up -d && sleep 30 && docker exec -i anodot-agent agent update` (sleep 30 -- to wait until streamsets is up)

scripts/agent.sh
```diff
@@ -0,0 +1,23 @@
+if [[ $1 == 'install' ]]; then
```
dariakharlan, line 1: instead of install call it 'upgrade'

dariakharlan: upgrade `docker-compose pull && docker-compose up -d && sleep 30 && docker exec -i anodot-agent agent update` (sleep 30 -- to wait until streamsets is up)

PR #62, opened 2020-05-28
Don't delete a pipeline if it already exists when creating using file

AntonZelenin:


agent/src/agent/pipeline/cli.py
```diff
@@ -105,7 +103,8 @@ def create(advanced, file):
 
     source_config_name = click.prompt('Choose source config', type=click.Choice(sources),
                                       default=get_default_source(sources))
-    pipeline_id = prompt_pipeline_id()
+    pipeline_id = click.prompt('Pipeline ID (must be unique)', type=click.STRING).strip()
```
dariakharlan, line 106: please keep it under `@infinite_retry` decorator so a user wouldn't have to run the command again in case the exception is raised

PR #69, opened 2020-06-04
Dev 1286 2 REST API for Anodot agent

AntonZelenin:


agent/src/agent/destination/http.py
```diff
@@ -67,81 +85,173 @@ def token(self):
...
+def create(
+    token: str,
+    url: str,
+    access_key: str = None,
+    proxy_host: str = None,
+    proxy_username: str = None,
+    proxy_password: str = None,
+    host_id: str = None,
+) -> Result[HttpDestination, str]:
+    return __build(HttpDestination.get_or_default(), token, url, access_key,
```
dariakharlan, line 168: Why when we create a destination you use `get_or_default`? Shouldn't it just create a new empty object?

agent/src/agent/destination/http.py
```diff
@@ -26,23 +33,34 @@ class HttpDestination:
...
+        self.host_id = self.generate_host_id()
+        self.access_key = ''
+
+    @staticmethod
+    def get() -> Optional['HttpDestination']:
+        if HttpDestination.exists():
+            dest = HttpDestination()
+            dest.__load()
+            return dest
+        return None
```
dariakharlan, line 47: I think an exception should be raised if it doesn't exist

agent/src/agent/destination/http.py
AntonZelenin: hmm, yes, it should

PR #73, opened 2020-06-16
dev-1213 sources REST API

AntonZelenin:


agent/src/agent/cli/source.py
```diff
@@ -60,7 +61,7 @@ def create_with_file(file):
             source_instance = source.create_object(item['name'], item['type'])
```
dariakharlan, line 61: maybe move `create_object` to the source repository?

agent/src/agent/cli/destination.py
```diff
@@ -79,14 +79,13 @@ def __start_monitoring_pipeline():
 
 @click.command()
 @click.option('-t', '--token', type=click.STRING, default=None)
-@click.option('--proxy/--no-proxy', default=False)
```
dariakharlan, line 82: Please don't remove it, you need to keep backward compatibility

PR #78, opened 2020-06-19
Created a script to create/update sources and pipelines

AntonZelenin:


agent/src/agent/pipeline/pipeline_manager.py
```diff
@@ -191,13 +199,19 @@ def enable_destination_logs(self, enable):
     def stop(self):
         stop_pipeline(self.pipeline.id)
 
-    def start(self):
+    def start(self) -> Result[bool, str]:
+        status = get_pipeline_status(self.pipeline.id)
+        if status == Pipeline.STATUS_RUNNING:
+            return Err(f'Pipeline {self.pipeline.id} is already running')
+        if status not in [Pipeline.STATUS_STOPPED, Pipeline.STATUS_EDITED]:
```
dariakharlan, line 206: raising an exception is ok in this case

PR #79, opened 2020-06-22
dev-1312 REST API for sources

AntonZelenin:


agent/Dockerfile.local
```diff
@@ -8,7 +8,7 @@ ADD .bashrc /root/.bashrc
 VOLUME /usr/src/app/data
 RUN apt update && apt install -y vim nano netcat curl nano
 
-ENV DATA_DIR="/data" STREAMSETS_USERNAME="admin" STREAMSETS_PASSWORD="admin" STREAMSETS_URL="http://localhost:18630" \
+ENV DATA_DIR="data" STREAMSETS_USERNAME="admin" STREAMSETS_PASSWORD="admin" STREAMSETS_URL="http://localhost:18630" \
```
dariakharlan, line 11: why?

agent/src/agent/api/routes/source.py
```diff
@@ -1,27 +1,50 @@
...
-@source.route('/source/<source_id>', methods=['DELETE'])
+    if not HttpDestination.exists():
+        return DESTINATION_DOESNT_EXIST, 400
+    try:
+        source.validate_json_for_create(request.get_json())
+        source_instances = []
+        for config in request.get_json():
+            source_instances.append(source.create_from_json(config).to_dict())
+    except Exception as e:
+        return str(e), 400
```
dariakharlan, line 26: Generic exceptions should return server error 500. Returning 400 bad request all the time may be confusing because it might not be a client's fault but something broken inside the code.

agent/src/agent/api/routes/source.py
```diff
@@ -1,27 +1,50 @@
...
+def edit():
+    if not HttpDestination.exists():
+        return DESTINATION_DOESNT_EXIST, 400
+    try:
+        source.validate_json_for_edit(request.get_json())
+        source_instances = []
+        for config in request.get_json():
+            source_instances.append(source.edit_using_json(config).to_dict())
+    except Exception as e:
+        return str(e), 400
```
dariakharlan, line 40: same here also, generic exceptions should return server error 500

agent/src/agent/api/routes/source.py
```diff
@@ -1,27 +1,50 @@
...
+    return jsonify(source_instances)
+
+
+@sources.route('/sources/<source_id>', methods=['DELETE'])
 def delete(source_id):
-    return 'Delete source'
+    try:
+        source_repository.delete_by_name(source_id)
+    except Exception as e:
+        return str(e), 400
```
dariakharlan, line 49: same here also, generic exceptions should return server error 500

agent/src/agent/validator/proxy.py
```diff
@@ -0,0 +1,11 @@
+import requests
+
+from agent import proxy
+
+
+def is_valid(proxy_obj: proxy.Proxy) -> bool:
+    try:
+        requests.get('http://example.com', proxies=proxy.get_config(proxy_obj), timeout=5)
```
dariakharlan, line 8: Can we check here `https://<destination_url>/api/v1/status`? Because some clients might whitelist only specific anodot ips

docker-compose-dev.yml
```diff
@@ -39,9 +39,9 @@ services:
       - dc
     volumes:
       - ./sdc-data:/sdc-data
-      - ./agent-data:/data
       - dummy-destination-data:/output
       - ./agent:/usr/src/app
+      - ./data:/usr/src/app/data
```
dariakharlan, line 44: Then in `Dockerfile.local` specify `DATA_DIR=/usr/src/app/data`

agent/src/agent/validator/proxy.py
AntonZelenin: we'll check in both places

docker-compose-dev.yml
AntonZelenin: as I reverted DATA_DIR to "data" we don't need it

agent/Dockerfile.local
AntonZelenin: it easier to test and debug when both docker container and local environments share same same destinations, sources and pipelines

PR #82, opened 2020-06-24
Source forms backup

AntonZelenin:


AntonZelenin:
For now this functionality doesn't suite our needs, we'll keep code agile and less strict

PR #103, opened 2020-07-17
Fixed populate_sources_and_pipelines.py script, added test

AntonZelenin:


agent/src/agent/cli/pipeline.py
```diff
@@ -51,15 +53,19 @@ def get_pipelines_ids_complete(ctx, args, incomplete):
...
+        pipelines = []
         for config in configs:
-            pipeline.manager.create_from_json(config)
+            pipelines.append(
+                pipeline.manager.create_from_json(config)
+            )
             click.secho(f'Created pipeline {config["pipeline_id"]}', fg='green')
     except (StreamSetsApiClientException, ValidationError, PipelineException) as e:
         raise click.ClickException(str(e))
+    return pipelines
```
dariakharlan, line 68: Why do you need to return something here? I don't see where you use it

agent/src/agent/cli/destination.py
```diff
@@ -78,6 +78,7 @@ def _start_monitoring_pipeline():
             pipeline_ = pipeline.manager.create_object('Monitoring', MONITORING_SOURCE_NAME)
             pipeline_manager = pipeline.manager.PipelineManager(pipeline_)
             click.secho('Starting Monitoring pipeline...')
+            # todo
```
dariakharlan, line 81: todo what?)

AntonZelenin: I'm going to change the next two lines that create directories for sources and pipelines, it shouldn't happen during destination creation

PR #115, opened 2020-08-13
Store destinations, sources and pipelines in a database

AntonZelenin:


agent/src/agent/scripts/migrate-to-db.py
```diff
@@ -0,0 +1,83 @@
...
+        if destination.repository.exists():
+            print('Destination already exists, skipping')
+            return
+        data = json.load(f)
+        dest = destination.HttpDestination()
+        dest.config = data['config']
+        dest.host_id = data['host_id']
+        dest.access_key = data['access_key']
+        destination.repository.create(dest)
+        print('Destination successfully migrated')
```
dariakharlan, line 34: why use print if you have a logger with stdout handler?

agent/src/agent/scripts/migrate-to-db.py
```diff
@@ -0,0 +1,83 @@
...
+def migrate_destination(data_dir):
+    with open(os.path.join(data_dir, 'destination.json')) as f:
+        if destination.repository.exists():
+            print('Destination already exists, skipping')
+            return
+        data = json.load(f)
+        dest = destination.HttpDestination()
+        dest.config = data['config']
+        dest.host_id = data['host_id']
+        dest.access_key = data['access_key']
```
dariakharlan, line 32: access key may not be present (if the agent version is very old)

agent/src/agent/scripts/migrate-to-db.py
```diff
@@ -0,0 +1,83 @@
...
+        print('Destination successfully migrated')
+
+
+def migrate_sources(data_dir):
+    if not source.repository.exists(MONITORING_SOURCE_NAME):
+        source.repository.create(source.Source(MONITORING_SOURCE_NAME, source.TYPE_MONITORING, {}))
+        print('Created monitoring source')
+    else:
+        print('Monitoring source already exists')
+    for root, _, filenames in os.walk(os.path.join(data_dir, 'sources')):
```
dariakharlan, line 43: sources directory always had only one level, no need to use `os.walk`, you can use `os.listdir` or `os.scandir`

dariakharlan: the same for pipelines

agent/src/agent/scripts/migrate-to-db.py
```diff
@@ -0,0 +1,83 @@
...
+        for filename in filenames:
```
dariakharlan, line 44: It's better to catch every exception in the loop, log it, and continue with the next file. Just in case something is broken in the user's env (which often is 😔). The same for pipelines

agent/src/agent/source/repository.py
```diff
@@ -1,78 +1,85 @@
...
+
+def get_by_name(source_name: str) -> source.Source:
+    return _construct_source(_get_entity(source_name))
+
+
+def _construct_source(source_entity: entity.Source) -> source.Source:
+    source_ = source.manager.create_source_obj(source_entity.name, source_entity.type)
+    source_.config = source_entity.config
+    source_.id = source_entity.id
+    # todo do we need this validation?
```
dariakharlan, line 70: You don't need it anymore, as all configs will be stored in the DB now

agent/src/agent/destination/repository.py
```diff
@@ -0,0 +1,49 @@
...
+def upsert(destination_: HttpDestination):
+    update(destination_) if exists() else create(destination_)
+
+
+def create(destination_: HttpDestination):
+    session.add(destination_.to_entity())
+    session.commit()
+
+
+def update(destination_: HttpDestination):
```
dariakharlan, line 26: For example, if you add a new column to the entity you will have to not forget to set it in two places, here and in the `to_entity` method. The same is for sources and pipelines. Maybe it's better to store the entity object inside the main object? Then `to_entity()` function will return a new entity if not exists and updated entity otherwise

agent/src/agent/source/source.py
```diff
@@ -1,14 +1,32 @@
...
     def __init__(self, name: str, source_type: str, config: dict):
         self.config = config
         self.type = source_type
         self.name = name
         self.sample_data = None
 
+    # todo refactor
+    def __getattr__(self, attr):
+        if attr == 'sample_data':
+            return None
```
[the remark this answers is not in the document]
AntonZelenin: no, when you run preview the data will be set into this attribute, so when you try to use it later this method won't be called

dariakharlan (approved):

PR #137, opened 2020-09-11
Fix single and multiple sources editing

AntonZelenin:


agent/src/agent/source/manager.py
```diff
@@ -51,15 +51,13 @@ def create_source_from_json(config: dict) -> source.Source:
 
 
 def edit_using_json(configs: dict) -> List[source.Source]:
-    validate_configs_for_edit(configs)
     exceptions = {}
     sources = []
     for config in configs:
```
dariakharlan, line 56: validate if list

dariakharlan (changes requested):

dariakharlan (approved):

PR #151, opened 2020-09-25
Fix pipeline status change in backup

AntonZelenin:


agent/src/agent/cli/backup.py
```diff
@@ -58,13 +59,20 @@ def _update_status(pipeline_: pipeline.Pipeline):
...
+                    pipeline.manager.stop(pipeline_)
+                except StreamSetsApiClientException:
+                    pipeline.manager.force_stop_pipeline(pipeline_.name)
             pipeline.manager.start(pipeline_)
 
 
-def _should_change_status(pipeline_: pipeline.Pipeline) -> bool:
-    return pipeline_.status != pipeline.manager.get_pipeline_status(pipeline_.name)
+def _should_change_status(pipeline_: Pipeline) -> bool:
+    # most likely we won't export broken pipelines, so only running, edited and stopped matter
```
dariakharlan, line 70: what do you mean most likely? what happens if we still do?
A pipeline can easily be in RETRY status for several minutes and a backup can be made during that time

dariakharlan (changes requested):

agent/src/agent/cli/backup.py
```diff
@@ -49,22 +51,21 @@ def _get_pipelines():
...
-            if not _can_start(pipeline_):
-                pipeline.manager.stop(pipeline_)
-            pipeline.manager.start(pipeline_)
-
-
-def _should_change_status(pipeline_: pipeline.Pipeline) -> bool:
-    return pipeline_.status != pipeline.manager.get_pipeline_status(pipeline_.name)
+def _update_status(pipeline_: Pipeline):
+    for function in _change_status_functions(pipeline_):
+        function(pipeline_)
```
dariakharlan, line 56: why do you need this, why don't just call the functions at once?

dariakharlan (changes requested):

agent/src/agent/cli/backup.py
AntonZelenin: dunno :)

dariakharlan (approved):

PR #152, opened 2020-09-30
Save influx offsets to streamsets

AntonZelenin:


agent/src/agent/pipeline/config/handlers/influx.py
```diff
@@ -156,28 +94,15 @@ def update_source_configs(self):
                     'dimensions': columns,
                     'metric': measurement_name,
                     'delay': delay,
-                    'interval': str(interval) + 's',
+                    # interval is in minutes
```
dariakharlan, line 97: why interval in minutes? don't change any config to not break backward compatibility

agent/src/agent/pipeline/config/jython_scripts/influx.py
```diff
@@ -0,0 +1,38 @@
...
+try:
+    sdc.importLock()
+    import time
+    from datetime import datetime, timedelta
+finally:
+    sdc.importUnlock()
+
+
+def get_now_with_delay():
+    return int(time.time()) - int(sdc.userParams['INTERVAL_IN_MINUTES']) * 60
```
dariakharlan, line 12: I don't see any delay here

dariakharlan (changes requested):

agent/src/agent/pipeline/config/handlers/influx.py
AntonZelenin: omg, I thought it was in minutes

agent/src/agent/pipeline/config/jython_scripts/influx.py
AntonZelenin: oops

dariakharlan (approved):

PR #159, opened 2020-10-08
Send agent pipelines to bc

AntonZelenin:


agent/src/agent/cli/destination.py
```diff
@@ -100,7 +100,6 @@ def _prompt_token(dest: HttpDestination):
 
 @infinite_retry
 def _prompt_access_key(dest: HttpDestination):
-    access_key = click.prompt('Anodot access key', type=click.STRING, default=dest.access_key or '')
-    if access_key and not agent.destination.validator.is_valid_access_key(access_key, dest.proxy, dest.url):
+    dest.access_key = click.prompt('Anodot access key', type=click.STRING, default=dest.access_key or '')
```
dariakharlan, line 103: no default value, access_key is required now

agent/src/agent/destination/repository.py
```diff
@@ -28,5 +28,16 @@ def delete():
     session().commit()
 
 
+def get_auth_token(destination_: HttpDestination) -> Optional[AuthenticationToken]:
+    if not destination_.id:
```
dariakharlan, line 32: why is it possible?

agent/src/agent/destination/repository.py
```diff
@@ -28,5 +28,16 @@ def delete():
...
+        return None
+    return session().query(AuthenticationToken).filter(AuthenticationToken.destination_id == destination_.id).first()
```
dariakharlan, line 34: why don't you have a relationship in the destination model and get the token by `destination_.auth_token` or something like that?

dariakharlan (changes requested):
Need to think about monitoring this also

agent/src/agent/destination/repository.py
AntonZelenin: because we need to validate access key while creating destination, by that moment destination's not saved to db yet. I wasn't able to come up with a better solution

agent/src/agent/destination/repository.py
AntonZelenin: I just was implementing it :) You'll see it in the next commit

agent/src/agent/destination/repository.py
AntonZelenin: hmm, as we'll use relationship it's not a problem anymore

agent/src/agent/destination/anodot_api_client.py
```diff
@@ -35,22 +36,29 @@ class AnodotApiClientException(click.ClickException):
...
         self.session = requests.Session()
-        self._get_auth_token()
-
-    def _get_auth_token(self):
-        auth_token = self.session.post(self._build_url('access-token'),
-                                       json={'refreshToken': self.access_key},
-                                       proxies=self.proxies)
-        auth_token.raise_for_status()
-        self.session.headers.update({'Authorization': 'Bearer ' + auth_token.text.replace('"', '')})
+        self.session.headers.update({'Authorization': 'Bearer ' + self._get_auth_token(destination_)})
```
dariakharlan, line 43: Need to check the token every time the request is made, not only on object creation

dariakharlan (changes requested):

agent/src/agent/scripts/send_to_bc.py
```diff
@@ -1,16 +1,39 @@
...
-    logger.error(traceback.format_exc())
+    send_error_metric()
+    raise
+
+for pipeline_ in pipelines:
+    if pipeline_.name != pipeline.MONITORING:
+        try:
+            api_client.send_pipeline_data_to_bc(pipeline.manager.transform_for_bc(pipeline_))
+        except Exception:
+            send_error_metric(pipeline_.name)
```
dariakharlan, line 38: I thought we agreed not to send an error metric in case of 404 response

dariakharlan (changes requested):

agent/src/agent/destination/destination.py
```diff
@@ -104,3 +108,30 @@ def get_proxy_url(self) -> str:
...
+    authentication_token = Column(String, nullable=False)
+    created_at = Column(TIMESTAMP(timezone=True), default=func.now())
+
+    destination = relationship('HttpDestination')
+
+    EXPIRATION_PERIOD_IN_SECONDS = 24 * 60 * 60
+
+    def __init__(self, destination_: HttpDestination, token: str):
+        self.destination_id = destination_.id
+        self.destination = destination_
```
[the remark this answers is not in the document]
AntonZelenin: good point

agent/src/agent/destination/anodot_api_client.py
```diff
@@ -0,0 +1,79 @@
...
+    pass
+
+
+class AnodotApiClient:
+    def __init__(self, destination_: HttpDestination, use_authentication: bool = True):
+        self.url = destination_.url
+        self.access_key = destination_.access_key
+        self.proxies = proxy.get_config(destination_.proxy)
+        self.session = requests.Session()
+        if use_authentication:
```
dariakharlan, line 40: I don't think you need this flag. Just in case of failed validation don't save anything to the db

dariakharlan (changes requested):

dariakharlan (approved):

PR #170, opened 2020-10-28
Changed creation of a message for pipeline info

AntonZelenin:


agent/src/agent/pipeline/info.py
```diff
@@ -67,11 +67,19 @@ def _get_timestamp(utc_time):
...
-        for i in pipeline_info['issues']['pipelineIssues']:
-            pipeline_issues.append('{severity} - {configGroup} - {configName} - {message}'.format(**i))
+        for issue in pipeline_info['issues']['pipelineIssues']:
+            pipeline_issues.append(_format(issue))
     return pipeline_issues
 
 
+def _format(info: dict) -> str:
+    keys = ['severity', 'configGroup', 'configName', 'message']
+    return ' - '.join(list(filter(
```
dariakharlan, line 77: `' - '.join([info[key] for key in keys if key in info])`
I think this is more understandable than list/filter/map

dariakharlan (changes requested):

dariakharlan (approved):

PR #172, opened 2020-10-29
Set pipeline id into agent offset url

AntonZelenin:


agent/src/agent/pipeline/config/stages/destination.py
```diff
@@ -5,7 +5,7 @@
 class Destination(Stage):
     def get_config(self) -> dict:
         return {
-            'conf.agentOffsetUrl': f'{AGENT_URL}/pipeline-offset/'
+            'conf.agentOffsetUrl': AGENT_URL + '/pipeline-offset/${PIPELINE_ID}'
```
dariakharlan, line 8: No need to set parameter,  you can use sdc function `${pipeline:id()}`

dariakharlan (changes requested):

dariakharlan (approved):

PR #179, opened 2020-11-18
[jdbc] fix so pipeline won't freeze if there is no data

dariakharlan:


scripts/docker-compose.yml
```diff
@@ -20,13 +20,16 @@ services:
...
+      - db
     environment:
       AGENT_DB_HOST: 'db'
       AGENT_DB_USER: 'agent'
       AGENT_DB_PASSWORD: 'agent'
       STREAMSETS_URL: 'http://dc:18630'
-#    ports:
-#    - 8080:80
+    ports:
+    - 8080:80
```
AntonZelenin, line 30: Spaces are missing :)
Why did you uncomment it?

dariakharlan: So user when installing the agent will have an agent api available

AntonZelenin (approved):

PR #181, opened 2020-11-19
Set pipeline offset when restoring from backup, delete offset on reset

AntonZelenin:


agent/src/agent/pipeline/manager.py
```diff
@@ -240,9 +240,23 @@ def update_source_pipelines(source_: Source):
         print(f'Pipeline {pipeline_.name} updated')
 
 
+def update_pipeline_offset(pipeline_: Pipeline):
+    offset = streamsets.manager.get_pipeline_offset(pipeline_)
+    if not offset:
+        return
```
dariakharlan, line 246: Maybe it's better to raise an exception here? Or at least write something to the logs

dariakharlan (changes requested):

agent/src/agent/pipeline/manager.py
AntonZelenin: It's not an exception, if you created a pipeline, didn't run it and called streamsets.manager.get_pipeline_offset(pipeline_) - it will return None.. or an empty string, so it's okay that offset might be empty

dariakharlan (approved):

PR #183, opened 2020-11-24
Monitoring prometheus compatible

dariakharlan:


agent/src/agent/monitoring/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from . import metrics, streamsets
+from agent.modules import constants, logger
+from datetime import datetime
+from prometheus_client import generate_latest
+
+logger_ = logger.get_logger(__name__)
+
+
+def get_latest():
+    streamsets.get_metrics()
```
AntonZelenin, line 12: could we give different names to functions like get_metrics()? It's very confusing to see a getter that doesn't return anything

agent/src/agent/streamsets/manager.py
```diff
@@ -29,23 +27,17 @@ def delete_streamsets(streamsets_: StreamSets):
...
 
 
 def _has_pipelines(streamsets_: StreamSets) -> bool:
-    pipelines = pipeline.repository.get_by_streamsets_id(streamsets_.id)
-    if len(pipelines) > 1:
-        return True
-    if len(pipelines) == 1 and not pipeline.manager.is_monitoring(pipelines[0]):
-        return True
-    return False
+    return True if len(pipeline.repository.get_by_streamsets_id(streamsets_.id)) >= 1 else False
```
AntonZelenin, line 34: `return len(pipeline.repository.get_by_streamsets_id(streamsets_.id)) > 0`

agent/src/agent/streamsets/manager.py
```diff
@@ -29,23 +27,17 @@ def delete_streamsets(streamsets_: StreamSets):
...
 
 
 def _can_move_pipelines():
     return len(streamsets.repository.get_all()) > 1
 
-
+# TODO: pass StreamSets instance as an argument
```
AntonZelenin, line 40: then we'll need to check pipeline_.streamsets in every place we call _client()
in order to avoid it we'll need to make major changes to streamsets manager design

agent/src/agent/streamsets/manager.py
```diff
@@ -139,7 +131,8 @@ def add_empty(s: StreamSets):
         if s.id not in pipeline_streamsets:
             pipeline_streamsets[s.id] = 0
     # choose streamsets with the lowest number of pipelines
-    pipeline_streamsets = pipeline.repository.count_by_streamsets()
+    pipeline_streamsets = streamsets.repository.count_pipelines_by_streamsets()
+    logger.info(pipeline_streamsets)
```
AntonZelenin, line 135: is this logger important? In my pr for sdc channel I removed the function count_by_streamsets()
and as I understand count_pipelines_by_streamsets() produces same result as count_by_streamsets()?

AntonZelenin (changes requested):

agent/src/agent/monitoring/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+    return generate_latest(registry=metrics.registry)
+
+
+def latest_to_anodot():
+    get_latest()
```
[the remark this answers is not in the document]
dariakharlan: to pull the latest metrics from streamsets

agent/src/agent/api/routes/monitoring.py
```diff
@@ -0,0 +1,69 @@
...
+    except destination.repository.DestinationNotExists:
+        return jsonify('')
+
+    data = monitoring_.latest_to_anodot()
+    monitoring_url = constants.MONITORING_URL if constants.MONITORING_URL else destination_.url
+
+    errors = []
+    if constants.MONITORING_SEND_TO_CLIENT:
+        url = urllib.parse.urljoin(monitoring_url,
+                                   f'api/v1/metrics?token={destination_.token}&protocol={destination_.PROTOCOL_20}')
```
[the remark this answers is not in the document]
dariakharlan: what exactly you want to put to the constant?

agent/src/agent/api/routes/monitoring.py
```diff
@@ -0,0 +1,69 @@
...
+        errors += _send_to_anodot(data, url, destination_.proxy)
+
+    if constants.MONITORING_SEND_TO_ANODOT:
+        url = urllib.parse.urljoin(monitoring_url,
+                                   f'api/v1/agents?token={destination_.token}&protocol={destination_.PROTOCOL_20}')
```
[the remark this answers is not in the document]
dariakharlan: what exactly you want to put to the constant?

agent/src/agent/streamsets/manager.py
dariakharlan: why does it matter?

agent/src/agent/monitoring/__init__.py
dariakharlan: pull_metrics is good enough?

agent/src/agent/streamsets/manager.py
dariakharlan: I think checking `if not pipeline_.streamsets:` is useless, agent will never have a pipeline without streamsets

agent/src/agent/streamsets/manager.py
dariakharlan: yeah, i'll remove the logger

the result of count_pipelines_by_streamsets is the same, the query is made differently.Now streamsets might be empty (because we don't have monitoring pipelines anymore) so we need to do `SELECT streamsets  OUTER JOIN pipelines`. The old query selects from pipelines table and skips streamsets instances which don't have pipelines yet

agent/src/agent/api/routes/monitoring.py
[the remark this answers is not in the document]
AntonZelenin: the whole url, same function call is used in two places. Or better create a function with parameters token and protocol

agent/src/agent/api/routes/monitoring.py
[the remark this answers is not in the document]
AntonZelenin: you can make the same function as above

agent/src/agent/monitoring/__init__.py
AntonZelenin: yep

agent/src/agent/streamsets/manager.py
AntonZelenin: it's just easier to read

agent/src/agent/streamsets/manager.py
AntonZelenin: when you create a new pipeline it doesn't have streamsets until you call streamsets.manager.create(pipeline). But there are situations when you balance streamsets and you want to create a pipeline in a specific instance, so you set it to pipeline before you pass it to create()
it's not the best solution but I wasn't able to come up with a better one yet

agent/src/agent/streamsets/manager.py
AntonZelenin: ok

agent/src/agent/api/routes/monitoring.py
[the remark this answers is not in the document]
dariakharlan: they are different urls - `api/v1/metrics` and `api/v1/agents`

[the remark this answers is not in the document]
dariakharlan: I don't see any point in creating a function here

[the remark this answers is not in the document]
AntonZelenin: search for api/v1/metrics in the project

agent/src/agent/streamsets/manager.py
dariakharlan: we can discuss it offline

agent/src/agent/monitoring/__init__.py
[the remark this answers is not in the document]
AntonZelenin: do you need get_latest() or only streamsets.get_metrics() here?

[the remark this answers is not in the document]
dariakharlan: get_latest() because I want all metrics to be sent to anodot, not only streamsets

AntonZelenin (approved):

PR #184, opened 2020-11-25
[antomation] - Fix deadlock

dariakharlan:


agent/alembic/versions/87dc1408358b_add_schema_column_to_pipelines.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('pipelines', sa.Column('schema', sa.JSON))
+
+
+def downgrade():
+    op.drop_column('streamsets', 'schema')
```
AntonZelenin, line 24: I guess you meant 'pipelines'?

agent/src/agent/pipeline/pipeline.py
```diff
@@ -68,8 +68,9 @@ class Pipeline(Entity):
     name = Column(String)
     source_id = Column(Integer, ForeignKey('sources.id'))
     destination_id = Column(Integer, ForeignKey('destinations.id'))
-    config = Column(MutableDict.as_mutable(JSON))
-    override_source = Column(MutableDict.as_mutable(JSON))
+    config = Column(JSON)
+    schema = Column(JSON)
+    override_source = Column(JSON)
```
AntonZelenin, line 73: The type of these fields must be MutableDict.as_mutable(JSON), otherwise, you won't be able to update their values

dariakharlan: We don't need to change the object itself. We can set a new value if we need to change something. Anyway, it's better not to change it at all.
Without these MutableDict there are no strange updates anymore so it fixes our deadlock problem.

AntonZelenin: then do we need to update other entities?

PR #186, opened 2020-11-30
Allow to specify a function as a value in kafka transofrmations

AntonZelenin:


agent/src/agent/pipeline/config/expression_parser/condition.py
```diff
@@ -2,9 +2,47 @@
 Validate condition entered by user and transform it to StreamSets expression language
 """
 import re
+import regex
 
 COMPARISON_FUNCTIONS = ['contains', 'startsWith', 'endsWith', 'matches']
 COMPARISON_LITERALS = ['==', '!=']
+# function name: num of arguments
+FUNCTIONS = {
+    'str:regExCapture': 3
```
dariakharlan, line 11: Can we easily add other functions that return string?

agent/src/agent/pipeline/config/expression_parser/condition.py
```diff
@@ -70,6 +108,35 @@ def validate(condition: str) -> bool:
     return True
 
 
+def validate_value(value: str):
```
dariakharlan, line 111: this is not used anywhere

agent/requirements.txt
```diff
@@ -55,3 +55,4 @@ werkzeug==0.16.1
 result==0.5.0
 wtforms-json==0.3.3
 alembic==1.4.2
+regex==2020.11.13
```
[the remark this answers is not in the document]
AntonZelenin: because re doesn't support recursion, I use it here to extract function arguments https://github.com/anodot/daria/blob/58389798ad6df28628be45d0d71b0124a65597ff/agent/src/agent/pipeline/config/expression_parser/condition.py#L170

agent/src/agent/pipeline/config/expression_parser/condition.py
AntonZelenin: well, yeah, just type a new function name here. But depending on a function we might need to change other code a bit, like if it takes a value from a record as a second argument, not the first

agent/src/agent/pipeline/config/expression_parser/condition.py
AntonZelenin: what do you mean?
https://github.com/anodot/daria/blob/6d95e42dc46b36219a4d5dba4dcfa0a1b57d36fc/agent/src/agent/pipeline/config/expression_parser/transformation.py#L11

agent/src/agent/pipeline/config/expression_parser/condition.py
dariakharlan: let's add the ones for which we don't need to change the code
str:replace

dariakharlan (approved):

PR #187, opened 2020-12-02
Send data about deleted pipelines to bc

AntonZelenin:


agent/src/agent/destination/anodot_api_client.py
```diff
@@ -50,27 +50,37 @@ def refresh_session_authorization(self):
...
 
     @endpoint
     def send_pipeline_data_to_bc(self, pipeline_data: dict):
-        return self.session.post(self._build_url('bc', 'agents'), proxies=self.proxies, json=pipeline_data)
+        return self.session.post(self._build_url(['bc', 'agents']), proxies=self.proxies, json=pipeline_data)
+
+    @endpoint
+    def delete_pipeline_from_bc(self, pipeline_id: str):
+        return self.session.delete(
+            self._build_url(['bc', 'agents'], {'pipelineId': pipeline_id}),
```
dariakharlan, line 84: No need to rewrite _build_url, pass parameters directly to the delete function
https://requests.readthedocs.io/en/master/user/quickstart/#passing-parameters-in-urls

agent/src/agent/pipeline/repository.py
```diff
@@ -68,3 +68,17 @@ def count_by_streamsets() -> Dict[int, int]:
 
 def get_by_streamsets_id(streamsets_id: int) -> List[Pipeline]:
     return session().query(Pipeline).filter(Pipeline.streamsets_id == streamsets_id).all()
+
+
+def add_deleted_pipeline_id(pipeline_id: str):
```
dariakharlan, line 73: Why you did not create an orm entity for deleted_pipelines table?

dariakharlan (changes requested):

agent/src/agent/pipeline/repository.py
AntonZelenin: just to keep it simple, I thought there's no need in creating a new class just to pass it's instances to the repository

dariakharlan (approved):

PR #193, opened 2020-12-11
Fixes for agent testing

dariakharlan:


agent/src/agent/cli/streamsets.py
```diff
@@ -27,10 +27,18 @@ def list_():
...
-    )
+@click.option('--url', type=click.STRING)
+@click.option('--username', type=click.STRING, default='admin')
+@click.option('--password', type=click.STRING, default='admin')
+@click.option('--agent-ext-url', type=click.STRING, default='http://anodot-agent')
+def add(url, username, password, agent_ext_url):
+    if url:
+        _validate_streamsets_url(url)
+        _validate_agent_external_url(agent_ext_url)
+        streamsets_ = StreamSets(url, username, password, agent_ext_url)
```
AntonZelenin, line 38: we also need to call streamsets.validator.validate(streamsets_) in order to check username and password, maybe it's possible to move this call from _prompt_streamsets() here?

agent/src/agent/cli/streamsets.py
```diff
@@ -27,10 +27,20 @@ def list_():
...
+        _validate_streamsets(streamsets_)
+    else:
+        streamsets_ = _prompt_streamsets(StreamSets(_prompt_url(), '', '', ''))
+        streamsets_.agent_external_url = _prompt_agent_external_url(streamsets_)
```
AntonZelenin, line 42: why don't you prompt it in the _prompt_streamsets() after streamsets validation?

dariakharlan: because if username/pass validation fails I want to do infinite retry
and if validation of agent_external_url fails i don't want to prompt username and pass again

AntonZelenin (approved):

PR #196, opened 2020-12-17
Move part of code into sdc-client library

AntonZelenin:


agent/src/agent/pipeline/pipeline.py
```diff
@@ -250,9 +252,23 @@ def delay(self) -> str:
     def batch_size(self) -> str:
         return self.config.get('batch_size', 1000)
 
-    def set_streamsets(self, streamsets):
-        self.streamsets_id = streamsets.id
-        self.streamsets = streamsets
+    def get_config(self) -> dict:
```
dariakharlan, line 255: This name `get_config` might be confusing because the pipeline entity has a property `config`

dariakharlan (changes requested):

agent/src/agent/pipeline/pipeline.py
AntonZelenin: ok, I'll rename it here as well as in the IPipeline in the lib, just didn't want to give a long name to the interface method

agent/src/agent/pipeline/manager.py
```diff
@@ -327,13 +330,23 @@ def get_sample_records(pipeline_: Pipeline) -> (list, list):
...
-        streamsets.manager.delete(test_pipeline)
+        sdc_client.delete(test_pipeline)
     return preview_data, errors
+
+
+def create_streamsets_pipeline_config(pipeline_: Pipeline) -> dict:
+    return get_config_handler(pipeline_).override_base_config()
+
+
+def _get_config_loader(pipeline_: Pipeline):
```
dariakharlan, line 349: Don't you have this method in agent/src/agent/pipeline/config/handlers/factory.py?

dariakharlan (changes requested):
agent/src/agent/pipeline/config/base_pipelines/Monitoring.json
agent/src/agent/pipeline/config/handlers/monitoring.py
these files should've been deleted, probably merge went wrong

Also you should remove all mentions of Monitoring pipeline from sdc-client

PR #197, opened 2020-12-17
Do not change config in the source after it's set

dariakharlan:


agent/src/agent/pipeline/manager.py
```diff
@@ -388,6 +388,6 @@ def _get_preview_data(test_pipeline: Pipeline):
     except (Exception, KeyboardInterrupt) as e:
         logger_.exception(str(e))
         raise
-    finally:
-        streamsets.manager.delete(test_pipeline)
+    # finally:
```
AntonZelenin, line 391: did you forget to uncomment?

agent/src/agent/pipeline/manager.py
```diff
@@ -388,6 +388,6 @@ def _get_preview_data(test_pipeline: Pipeline):
...
+    #     streamsets.manager.delete(test_pipeline)
```
AntonZelenin, line 392: did you forget to uncomment?

agent/src/agent/pipeline/pipeline.py
```diff
@@ -101,8 +101,8 @@ def config_changed(self) -> bool:
     def set_config(self, config: dict):
         self._previous_config = deepcopy(self.config)
         self._previous_override_source = deepcopy(self.override_source)
-        self.override_source = config.pop(self.OVERRIDE_SOURCE, {})
-        self.config = config
+        self.config = deepcopy(config)
+        self.override_source = self.config.pop(self.OVERRIDE_SOURCE, {})
```
AntonZelenin, line 105: why do you make a deep copy here and why don't you pop override source from it first?

agent/tests/api/source/test_elastic.py
```diff
@@ -25,7 +25,7 @@ class TestElastic:
                     'conf.queryInterval': 'seconds',
                 }
             }],
-            'er': b'[{"config":{"conf.httpUris":["http://es:9201"],"conf.index":"test1","conf.initialOffset":"now-3d/d","conf.isIncrementalMode":false,"conf.offsetField":"timestamp","conf.query":null,"conf.queryInterval":"seconds"},"name":"elastic","type":"elastic"}]\n'
+            'er': b'[{"config":{"conf.httpUris":["http://es:9201"],"conf.index":"test1","conf.initialOffset":"now-3d/d","conf.isIncrementalMode":false,"conf.offsetField":"timestamp","conf.queryInterval":"seconds"},"name":"elastic","type":"elastic"}]\n'
```
AntonZelenin, line 28: could you please change the string here to array and compare it to result.json instead of result.data?

agent/src/agent/pipeline/manager.py
dariakharlan: yes :)

agent/src/agent/pipeline/pipeline.py
dariakharlan: Because `config.pop(self.OVERRIDE_SOURCE, {})` changes the `config` object and we need to create a pipeline with this object two times - for preview and later for actual pipeline creation. So after we create a test pipeline `config` object doesn't have `self.OVERRIDE_SOURCE` property anymore and we need it to create the pipeline and save it to db

AntonZelenin: got it, now I see you pop from self.config so that's perfectly fine :)

agent/src/agent/cli/prompt/pipeline/elastic.py
```diff
@@ -29,4 +29,8 @@ def prompt_query(self):
             errors = elastic_query.get_errors(query, offset_field)
             if errors:
                 raise click.ClickException(errors)
-            self.config['query'] = query
+
+            # todo duplicate
+            if 'override_source' not in self.config:
+                self.config['override_source'] = {}
```
AntonZelenin, line 35: let's create it in the construct of PromptConfig and also remove from def prompt()

agent/src/agent/pipeline/config/stages/base.py
```diff
@@ -13,13 +13,7 @@ class Stage(ABC):
...
         self.pipeline = pipeline_
         self.sdc_stage = sdc_stage
-        if isinstance(pipeline_, pipeline.TestPipeline):
-            self.config = self._get_source_config()
-        else:
-            self.config = self._get_config()
-
-    def _get_source_config(self):
-        return {**self.pipeline.source.config, **self.pipeline.override_source}
+        self.config = self._get_config()
```
AntonZelenin, line 16: please remove unused import in this file

AntonZelenin (approved):

PR #200, opened 2021-01-15
Dev 1647 - protocol 3 for jdbc

AntonZelenin:


AntonZelenin:
TODO: update docs, update agent upgrade instruction

agent/tests/test_pipelines/test_directory_http.py
```diff
@@ -47,12 +47,18 @@ def test_stop(self, cli_runner, name):
...
-        schema_id = '111111-22222-3333-4444'
-        assert get_output(f'{schema_id}_watermark.json') == {'watermark': 1512892800.0, 'schemaId': schema_id}
+        schema_id = get_schema_id('test_dir_json')
+        assert get_output(f'{schema_id}_watermark.json') == {'watermark':   1512892800.0, 'schemaId': schema_id}
 
     def test_force_stop(self, cli_runner, name=None):
         pytest.skip()
 
+    def test_output(self, name, pipeline_type, output):
+        expected_output = get_expected_output(name, output, pipeline_type)
```
dariakharlan, line 57: use get_expected_schema_output method?

agent/src/agent/pipeline/manager.py
```diff
@@ -318,7 +322,7 @@ def transform_for_bc(pipeline_: Pipeline) -> dict:
         'progress': {
             'last_offset': pipeline_.offset.offset if pipeline_.offset else '',
         },
-        'schema': pipeline_.get_schema() if pipeline_.get_schema_id() else schema.build(pipeline_),
+        'schema': pipeline_.get_schema(),
```
dariakharlan, line 325: what about pipelines that don't use schema yet?

agent/src/agent/pipeline/manager.py
```diff
@@ -23,17 +23,20 @@
...
-    def load_config(self, config, edit=False):
-        self.pipeline.set_config(self.file_loader.load(config, edit))
-        self._validate_config()
-
-    def _validate_config(self):
-        get_config_validator(self.pipeline).validate(self.pipeline)
+def get_default_protocol(pipeline_: Pipeline):
+    if pipeline_.protocol:
+        return pipeline_.protocol
+    else:
```
dariakharlan, line 29: you don't need to write `else` if you have a `return` at the end of `if`
Code with fewer nesting levels is easier to read

agent/src/agent/cli/prompt/pipeline/directory.py
```diff
@@ -9,6 +9,7 @@ def prompt_config(self):
             raise click.UsageError('No api key configured. Please configure api key using `agent destination` command')
         super().prompt_config()
         self.prompt_flush_bucket_size()
+        self.pipeline.protocol = self.pipeline.destination.PROTOCOL_30
```
dariakharlan, line 12: is it set when you create the pipeline with a file?

agent/alembic/versions/669a641624ed_add_protocol_column_to_pipelines.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '669a641624ed'
+down_revision = '63aac660c5f0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column('pipelines', sa.Column('protocol', sa.String))
```
dariakharlan, line 20: I think using a separate column and a property protocol is redundant. `use_schema` property in the config should be enough. And it will be easier to get rid of this when all pipelines will always use protocol 3.0

dariakharlan (changes requested):

agent/tests/test_pipelines/test_mysql_http.py
```diff
@@ -31,3 +44,26 @@ def test_info(self, cli_runner, name=None):
...
+        super().test_create_with_file(cli_runner, file_name)
+
+    def test_start(self, cli_runner, name):
+        super().test_start(cli_runner, name)
+
+    def test_force_stop(self, cli_runner, name):
+        super().test_force_stop(cli_runner, name)
+
+    def test_output(self, name, pipeline_type, output):
+        if name == 'test_mysql_no_schema':
```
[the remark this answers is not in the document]
AntonZelenin: will create

agent/src/agent/cli/prompt/pipeline/kafka.py
```diff
@@ -21,8 +21,8 @@ def prompt_config(self):
         self.transform()
 
     def set_consumer_group(self):
-        self.config['override_source'][source.KafkaSource.CONFIG_CONSUMER_GROUP] =\
-            click.prompt('Consumer group name', self._get_default_consumer_group())
+        consumer_group = click.prompt('Consumer group name', self._get_default_consumer_group())
+        self.pipeline.override_source = {source.KafkaSource.CONFIG_CONSUMER_GROUP: consumer_group}
```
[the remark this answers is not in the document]
AntonZelenin: you will write self.pipeline.override_source[key] = value? Don't see a problem to do that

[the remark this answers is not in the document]
AntonZelenin: or you mean to initialize it first?

agent/src/agent/pipeline/manager.py
```diff
@@ -169,35 +156,52 @@ def edit_using_json(configs: list) -> List[Pipeline]:
...
-            )
+            pipelines.append(edit_pipeline_using_json(config))
         except Exception as e:
             exceptions[config['pipeline_id']] = f'{type(e).__name__}: {str(e)}'
         if exceptions:
             raise pipeline.PipelineException(json.dumps(exceptions))
     return pipelines
 
 
+class PipelineWatcher:
```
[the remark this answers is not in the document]
AntonZelenin: yes. There are two places now. Ideally, we should have one place where that check will happen.
To have a method Pipeline.config_changed() is not good either, it adds one more responsibility to Pipeline and you can call that method anytime but it will work only when you called set_config first
When I changed code a bit I got an error when calling config_changed() because when you get a pipeline from the db it doesn't have _previous_config, nor _orevious_override_source. That's why I removed it

agent/src/agent/pipeline/config/schema.py
```diff
@@ -2,7 +2,24 @@
...
 
 
-def build(pipeline: Pipeline) -> dict:
+def create(pipeline: Pipeline):
+    api_client = anodot_api_client.AnodotApiClient(pipeline.destination)
+    created_schema = api_client.create_schema(_build(pipeline))
+    return created_schema['schema']
+
+
+def update(pipeline: Pipeline) -> dict:
```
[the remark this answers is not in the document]
AntonZelenin: hmm, you're right

agent/src/agent/pipeline/manager.py
AntonZelenin: it will return empty dict. I guess we don't need to send schema for pipelines that don't actually use it

agent/src/agent/pipeline/manager.py
AntonZelenin: yep

agent/alembic/versions/669a641624ed_add_protocol_column_to_pipelines.py
AntonZelenin: I created a separate column because it's not a part of pipeline configuration, it's rather a flag that shows what pipeline type we should build.
when all pipelines use protocol 3 we'll just drop that column.

agent/src/agent/cli/prompt/pipeline/directory.py
AntonZelenin: yes, it's set when we load json config into a pipeline

agent/src/agent/pipeline/manager.py
dariakharlan: it's the agent in bc script, the logic was that we build schema even if the pipeline is not using it. this way anodot UI can show dimensions and measurements

agent/src/agent/pipeline/manager.py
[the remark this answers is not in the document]
dariakharlan: why not just fix the config_changed method then? If set_config wasn't called and we just queried the object from the db then we should just return False, right?

agent/src/agent/cli/prompt/pipeline/kafka.py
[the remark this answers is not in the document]
dariakharlan: yes, initialize it first. If I want to add something to override_source and it happens to be called first then this consumer group setting will just override everything

agent/src/agent/pipeline/manager.py
AntonZelenin: really? Didn't know it, I thought that's a bug

dariakharlan (approved):

PR #210, opened 2021-02-10
Dev 1737 zabbix query by hosts

dariakharlan:


agent/src/agent/pipeline/config/jython_scripts/zabbix.py
```diff
@@ -14,7 +14,8 @@
 
 entityName = ''
 N_REQUESTS_TRIES = 3
-BATCH_SIZE = 1000
+ITEMS_BATCH_SIZE = 1000
+HISTORIES_BATCH_SIZE = 10
```
dariakharlan, line 18: set batch sizes as pipeline configs

agent/src/agent/pipeline/config/jython_scripts/zabbix.py
```diff
@@ -74,57 +79,140 @@ def get_interval():
...
+
+
+def fetch_hosts():
+    res = client.post('host.get', {'output': ['hostid', 'name']})
+    if len(res) == 0:
+        sdc.log.info('host.get - No data')
+    hosts = {}
+    for host in res:
+        hosts[host['hostid']] = host
+    return hosts
```
dariakharlan, line 127: `return {host['hostid']: host for host in res}`

this is not necessary, but just in case you didn't know :)

agent/src/agent/pipeline/config/jython_scripts/zabbix.py
```diff
@@ -74,57 +79,140 @@ def get_interval():
...
+        ))
+        end -= interval
+    return itemids_value_types
+
+
+def fetch_items_data(itemids):
+    global hosts
+    start = time.time()
+    items = {}
+    for item in client.post('item.get', {'itemids': itemids}):
```
dariakharlan, line 154: if a client in his query specified output parameter - set it here also

dariakharlan (commented):
logs are too noisy, set debug level or remove the most frequent ones

agent/src/agent/pipeline/config/jython_scripts/zabbix.py
AntonZelenin: yep, thanks, I was too lazy to search for this solution)

PR #246, opened 2021-04-20
Change how metrics are pulled from Cacti

AntonZelenin:


agent/src/agent/data_extractor/cacti/repository.py
```diff
@@ -10,98 +10,136 @@ def _get_session(connection_string: str):
...
+
+    def _get_graph_variables(self):
+        res = self.session.execute("""
+            select gl.id, field_name, field_value
+            from host_snmp_cache hsc
+            join graph_local gl
+            on gl.host_id = hsc.host_id
+            and gl.snmp_index = hsc.snmp_index
+            and gl.snmp_query_id = hsc.snmp_query_id
+        """)
```
dariakharlan, line 77: How long does this query execute? Maybe we need to force index here also

agent/src/agent/data_extractor/cacti/repository.py
```diff
@@ -10,98 +10,136 @@ def _get_session(connection_string: str):
...
+            local_graph_id = row['local_graph_id']
+            if local_graph_id not in self.graphs:
+                continue
+            if 'items' not in self.graphs[local_graph_id]:
+                self.graphs[local_graph_id]['items'] = []
+            if row['item_id'] not in self.graphs[local_graph_id]['items']:
+                continue
+
+            self.graphs[local_graph_id]['items'][row['item_id']]['data_source_name'] = row['data_source_name']
+            self.graphs[local_graph_id]['items'][row['item_id']]['item_title'] = row['item_title']
```
dariakharlan, line 106: _get_sources query and this one are the same, maybe we can join data_local here and get the rrd path?

dariakharlan (changes requested):
- Set up a cronjob in `uwsgi.ini` file for caching once in 30 mins
- Add a new metric to monitoring for measuring cronjobs execution time

agent/src/agent/pipeline/repository.py
```diff
@@ -30,6 +30,10 @@ def get_by_id(pipeline_id: str) -> Pipeline:
     return pipeline_
 
 
+def get_by_type(type_: str) -> List[Pipeline]:
+    return list(filter(lambda x: x.source.type == type_, get_all()))
```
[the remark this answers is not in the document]
AntonZelenin: I think it doesn't matter, even if there will be several hundred pipelines it will still work pretty fast

agent/src/agent/data_extractor/cacti/repository.py
AntonZelenin: I didn't notice any problems with this query, when I run it in the console it takes about 3 seconds

dariakharlan: If you mean console in pycharm then it selects only 500 rows this is why it works fast there. If you check with `count` statement it takes 50 seconds. With forcing index `count` takes 3,5 seconds. So I think you should add it

dariakharlan (approved):

PR #250, opened 2021-05-06
save transform value to the db so we don't need to keep the file

dariakharlan:


agent/src/agent/pipeline/json_builder.py
```diff
@@ -150,6 +150,17 @@ def load(self, client_config) -> dict:
 
         jsonschema.validate(self.client_config, schema)
         client_config.pop('source', None)
+
+        condition = self.client_config.get('filter', {}).get('condition')
+        if condition:
+            expression_parser.condition.validate(condition)
+
+        transformation = self.client_config.get('transform', {}).get('file')
```
AntonZelenin, line 158: please call the variable `transformation_file`, then it's more clear what it contains

AntonZelenin (approved):

PR #257, opened 2021-05-20
Added influx schema pipelines

AntonZelenin:


agent/src/agent/pipeline/config/handlers/influx.py
```diff
@@ -67,3 +65,12 @@ def get_optional_dimensions(self):
     @staticmethod
     def replace_illegal_chars(string: str) -> str:
         return string.replace(' ', '_').replace('.', '_').replace('<', '_')
+
+
+class InfluxSchemaConfigHandler(SchemaConfigHandler, InfluxConfigHandler):
```
dariakharlan, line 70: where is `'JavaScriptEvaluator_01': stages.js_convert_metrics.JSConvertMetrics30,`? 
why do you need to inherit from InfluxConfigHandler?

agent/src/agent/pipeline/config/stages/expression_evaluator.py
```diff
@@ -43,11 +43,14 @@ def _get_config(self) -> dict:
         expressions = []
         for key, val in self.pipeline.static_dimensions.items():
             expressions.append(get_value(self._get_dimension_field_path(key), f'"{val}"'))
-        timestamp_to_unix = get_convert_timestamp_to_unix_expression(self.pipeline.timestamp_type,
-                                                                     "record:value('/timestamp')",
-                                                                     self.pipeline.timestamp_format,
-                                                                     self.pipeline.timezone)
-        expressions.append(get_value('/timestamp', timestamp_to_unix))
+        if 'timestamp' in self.pipeline.config:
```
dariakharlan, line 46: for every influx pipeline we can set timestamp type as unix_ms and field name `time`

dariakharlan (changes requested):
Let's write an upgrade script that updates existing pipelines and sets regular values format so we don't need to support if conditions in multiple places. The goal is for influx pipelines to use the same configs as the others

agent/src/agent/pipeline/config/handlers/influx.py
AntonZelenin: why do you need to inherit from InfluxConfigHandler? - because its _override_stages() method adds required fields to the js evaluator and sets values to initScript

dariakharlan (changes requested):

dariakharlan (approved):

PR #259, opened 2021-05-31
Cacti totals

dariakharlan:


agent/src/agent/data_extractor/cacti/cacti.py
```diff
@@ -48,35 +60,83 @@ def extract_metrics(pipeline_: Pipeline, start: str, end: str, step: str) -> lis
...
+            continue
+        result[key] = deepcopy(metrics_to_sum[metric['properties']['what']])
+        result[key]['properties']['what'] = metric['properties']['what']
+        result[key]['value'] = metric['value']
+        result[key]['timestamp'] = metric['timestamp']
+
+    return list(result.values())
+
+
+def _get_values_for_item(
```
AntonZelenin, line 97: I think the name _build_metrics_for_item() or something like this would be better. I mean to use the word `metrics` instead of `values`

agent/src/agent/data_extractor/cacti/cacti.py
```diff
@@ -48,35 +60,83 @@ def extract_metrics(pipeline_: Pipeline, start: str, end: str, step: str) -> lis
...
-                    metric['value'] = value
-                    metric['timestamp'] = timestamp
-                    metrics.append(metric)
+            graph_metrics += _get_values_for_item(
+                result,
+                item,
+                start,
+                end,
+                step,
+                should_convert_to_bits and pipeline_.config['convert_bytes_into_bits'],
```
AntonZelenin, line 69: maybe lets put pipeline_.config['convert_bytes_into_bits'] into _should_convert_to_bits()?
And you can call it directly here without creating a variable

AntonZelenin (approved):

AntonZelenin (approved):

PR #273, opened 2021-07-19
Snmp agent

AntonZelenin:


agent/src/agent/pipeline/config/jython_scripts/snmp.py
```diff
@@ -0,0 +1,71 @@
...
+
+while True:
+    if sdc.isStopped():
+        break
+    if offset > get_now():
+        time.sleep(offset - get_now())
+
+    batch = sdc.createBatch()
+
+    res = requests.get(sdc.userParams['SNMP_SOURCE_URL'])
```
dariakharlan, line 52: please add timeout just in case

agent/src/agent/pipeline/config/jython_scripts/snmp.py
```diff
@@ -0,0 +1,71 @@
...
+        metric['timestamp'] = offset
+        record = sdc.createRecord('record created ' + str(datetime.now()))
+        record.value = metric
+        batch.add(record)
+
+        if batch.size() == sdc.batchSize:
+            batch.process(entityName, str(offset))
+            batch = sdc.createBatch()
+
+    event = sdc.createEvent('sage_error', 1)
```
dariakharlan, line 64: this is not sage

agent/src/agent/data_extractor/snmp/snmp.py
```diff
@@ -0,0 +1,83 @@
...
+        error_indication, error_status, error_index, var_binds = response
+        if error_indication:
+            logger_.error(error_indication)
+            continue
+        elif error_status:
+            logger_.error('%s at %s' % (
+                error_status.prettyPrint(),
+                error_index and var_binds[int(error_index) - 1][0] or '?'
+            ))
+            continue
```
dariakharlan, line 40: shouldn't we send some kind of error in the response? code 500, for example. Or send a metric to monitoring. Otherwise we won't be able to know if there were any errors we'll just get no data

dariakharlan (changes requested):

dariakharlan (approved):

PR #276, opened 2021-07-29
Influx 2.0

AntonZelenin:


AntonZelenin:
Need to add timestamp conversion to float

agent/src/agent/pipeline/config/jython_scripts/influx2_transform_records.py
```diff
@@ -0,0 +1,41 @@
...
+    if timestamp in records:
+        r = records[timestamp]
+        r[record.value['_field']] = record.value['_value']
+        continue
+    record.value[record.value['_field']] = record.value['_value']
+    record.value[sdc.userParams['TIMESTAMP_COLUMN']] = timestamp
+    records[timestamp] = record.value
+
+records = sorted(records.items(), key=lambda x: x[1][sdc.userParams['TIMESTAMP_COLUMN']])
+records = collections.OrderedDict(records)
```
dariakharlan, line 36: why do you need to convert it to an ordered dict?

agent/src/agent/pipeline/config/jython_scripts/influx2_transform_records.py
```diff
@@ -0,0 +1,41 @@
...
+
+def to_timestamp(date):
+    epoch = datetime(1970, 1, 1)
+    return (date - epoch).total_seconds()
+
+
+records = {}
+for record in sdc.records:
+    # * 1000 because of strange implementaion of 'Add tags and static dimensions', it will divide by 1000 later
+    timestamp = to_timestamp(datetime.strptime(record.value[sdc.userParams['TIMESTAMP_COLUMN']], DATEFORMAT)) * 1000
```
dariakharlan, line 26: if you don't want it to be divided by 1000 then specify timestamp format as `unix` instead of `unix_ms`

agent/src/agent/pipeline/config/jython_scripts/influx2_transform_records.py
```diff
@@ -0,0 +1,41 @@
...
+
+DATEFORMAT = '%Y-%m-%dT%H:%M:%S.%fZ'
+
+
+def to_timestamp(date):
+    epoch = datetime(1970, 1, 1)
+    return (date - epoch).total_seconds()
+
+
+records = {}
```
dariakharlan, line 23: why do you need a dict here and all those manipulations?
I think just doing `record.value[record.value['_field']] = record.value['_value']` would be enough

agent/src/agent/pipeline/config/stages/source/influx2.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Influx2Source(InfluxScript):
+    JYTHON_SCRIPT = 'influx2.py'
+
+    def get_config(self) -> dict:
+        config = super().get_config()
+        config['scriptConf.params'].extend([
+            {'key': 'URL', 'value': self._get_url()},
+            {'key': 'HEADERS', 'value': self._get_headers()},
+            {'key': 'DATA', 'value': self._get_data()},
```
dariakharlan, line 13: 1. I would call it query instead of data
2. What about filtering conditions?
3. Maybe it's better to allow the user to write his own queries as we do for victoria agents etc.?

dariakharlan (changes requested):

agent/src/agent/pipeline/config/jython_scripts/influx2_transform_records.py
AntonZelenin: need to change transform

dariakharlan (approved):

PR #280, opened 2021-09-02
Count error statuses end send notification to bc

AntonZelenin:


agent/src/agent/pipeline/manager.py
```diff
@@ -207,6 +228,13 @@ def transform_for_bc(pipeline_: Pipeline) -> dict:
     return data
 
 
+def should_send_error_notification(pipeline_: Pipeline) -> bool:
+    retries = pipeline.repository.get_pipeline_retries(pipeline_)
```
dariakharlan, line 231: querying pipeline retries each time for every pipeline separately is not effective. You should use joins when querying all pipelines in the bc script

dariakharlan (changes requested):
Please add an environmental variable to disable notifications globally also

dariakharlan (approved):

PR #285, opened 2021-09-15
Send watermark by cron

AntonZelenin:


agent/src/agent/scripts/send_watermarks.py
```diff
@@ -0,0 +1,64 @@
...
+
+    num_of_errors = 0
+    for pipeline_ in pipelines:
+        if pipeline_.uses_schema \
+                and pipeline_.periodic_watermark_config \
+                and pipeline_.offset.timestamp + pipeline_.watermark_delay <= time.time():
+            try:
+                api_client.send_watermark_to_bc(pipeline_.get_schema_id(), _get_next_bucket_start(pipeline_))
+            except requests.HTTPError as e:
+                if not e.response.status_code == 404:
```
dariakharlan, line 53: no need for silencing 404 here, this endpoint should be available everywhere

dariakharlan (changes requested):

dariakharlan (approved):

PR #287, opened 2021-09-24
oracle

dariakharlan:


agent/src/agent/source/validator.py
```diff
@@ -147,6 +150,24 @@ def validate_connection_string(self):
             raise ValidationException('Wrong url scheme. Use `clickhouse`')
 
 
+class OracleValidator(Validator):
+    VALIDATION_SCHEMA_FILE = 'jdbc.json'
+
+    def validate(self):
+        self.validate_json()
+        self.validate_connection_string()
+        self.validate_connection()
```
AntonZelenin, line 159: you can inherit JDBCValidator and remove VALIDATION_SCHEMA_FILE and def validate(self):

agent/src/agent/source/validator.py
```diff
@@ -147,6 +150,24 @@ def validate_connection_string(self):
...
+
+    @if_validation_enabled
+    def validate_connection_string(self):
+        url = deepcopy(self.source.config[source.JDBCSource.CONFIG_CONNECTION_STRING])
```
AntonZelenin, line 163: if url is a string you don't need deepcopy here

agent/tests/test_input/test_oracle.py
```diff
@@ -0,0 +1,26 @@
...
+        'test_create_with_file': [{'file_name': 'jdbc_pipelines_oracle'}],
+    }
+
+    def test_source_create(self, cli_runner):
+        pass
+
+    def test_create(self, cli_runner):
+        pass
+
+    def test_create_advanced(self, cli_runner):
```
AntonZelenin, line 18: you don't need these methods, the base class has only "with_file" methods

AntonZelenin (approved):

PR #307, opened 2021-11-12
Observium fixes

AntonZelenin:


agent/src/agent/pipeline/pipeline.py
```diff
@@ -132,47 +128,51 @@ def flush_bucket_size(self, value: str):
...
+    def all_dimensions(self) -> list:
+        if not self.dimensions or type(self.dimensions) is list:
+            return self.dimensions + self.static_dimension_names
+        return self.required_dimensions + self.optional_dimensions + self.static_dimension_names
 
     @property
-    def required_dimensions_paths(self) -> list:
-        return [self._get_property_path(value) for value in self.required_dimensions]
+    def dimension_names(self) -> list:
+        # todo why replace / ? are you sure we need to replace it?
```
dariakharlan, line 170: when you have a nested json, user specifies a path to the property with `level1/level2/level3`. For some reason, I decided that we need to replace '/' with '_' so we can't change it because it will break metrics for old pipelines

agent/src/agent/pipeline/pipeline.py
```diff
@@ -195,30 +195,45 @@ def values(self) -> dict:
...
     @property
     def target_types(self) -> list:
         if self.source.type == source.TYPE_INFLUX:
-            return [self.config.get('target_type', 'gauge')] * len(self.value_names)
-        return list(self.config['values'].values())
+            return [self.config.get('target_type', 'gauge')] * len(self.value_paths)
+        return list(self.values.values())
+
+    @property
+    def measurement_paths_with_names(self) -> dict:
```
dariakharlan, line 208: why we need this?

agent/src/agent/pipeline/config/js_scripts/protocol_3.js
```diff
@@ -19,17 +19,13 @@ function replace_illegal_chars(str) {
...
     var dimensions = {};
-    for (var j = 0; j < state['DIMENSIONS'].length; j++) {
-        var dimension = extract_value(record.value, replace_illegal_chars(state['DIMENSIONS'][j]))
-        if (dimension === null) {
-            continue;
-        }
-        dimension = String(dimension).trim()
-        if (dimension === '') {
-            continue;
+    for (var dimension_path in state['DIMENSIONS']) {
```
dariakharlan, line 22: a client specifies in the config 'new name': 'original path', but here its the other way around, I think it's confusing, in the client config you should also have 'original path': 'new name'

agent/src/agent/pipeline/config/js_scripts/protocol_3.js
```diff
@@ -19,17 +19,13 @@ function replace_illegal_chars(str) {
...
+        var dimension_value = extract_value(record.value, replace_illegal_chars(dimension_path))
+        dimension_value = String(dimension_value).trim()
+        if (dimension_value !== '' && dimension_value !== 'null') {
```
dariakharlan, line 25: revert to the original logic because value can be a 'null' string and it's valid

dariakharlan (changes requested):

agent/src/agent/pipeline/config/stages/js_convert_metrics.py
```diff
@@ -5,9 +5,9 @@ class JSConvertMetrics(base.Stage):
     def get_js_vars(self):
         return f"""
 state['TIMESTAMP_COLUMN'] = '{self.pipeline.timestamp_path}';
-state['DIMENSIONS'] = {self.pipeline.dimensions_paths};
-state['DIMENSIONS_NAMES'] = {self.pipeline.dimensions_names};
-state['VALUES_COLUMNS'] = {self.pipeline.values_paths};
+state['DIMENSIONS'] = {self.pipeline.dimension_paths};
```
dariakharlan, line 8: why not plural? it sounds like it's multiple paths of a single dimension

dariakharlan (changes requested):

agent/src/agent/pipeline/config/handlers/base.py
```diff
@@ -96,6 +96,10 @@ class BaseConfigHandler:
     def __init__(self, pipeline: Pipeline, base_config: dict):
         self.config = base_config
         self.pipeline = pipeline
+        self._check_pipeline()
```
dariakharlan, line 99: maybe we also need to rename this class because base sounds like it fits for everything

dariakharlan (changes requested):

agent/src/agent/pipeline/pipeline.py
AntonZelenin: well, for convenience. With this we don't need to use for loops with indices, we can use this method

agent/src/agent/pipeline/config/js_scripts/protocol_3.js
AntonZelenin: but logic didn't change, I just removed the continue operator and put code into the if block

agent/src/agent/pipeline/config/stages/js_convert_metrics.py
AntonZelenin: in English it's correct to say `dimension paths, I checked it with Brian

agent/src/agent/pipeline/config/handlers/base.py
AntonZelenin: probably, but it will affect lots of files, also we have another class BaseConfigLoader which in this case also should be renamed. Are you okay with this?

dariakharlan (approved):

PR #313, opened 2021-11-25
Fix snmp offset calculation

AntonZelenin:


AntonZelenin:
`to_timestamp()` and `get_now()` functions return different values, one returns as local time, another as UTC time. Also watermark in SNMP currently makes no sense as it's the same as last value

AntonZelenin:
Currently, in all Jython scripts where the `strptime()` is used, it thinks that provided datetime is in UTC and doesn't support timezones

dariakharlan (approved):

PR #318, opened 2021-12-01
Added style and config files for pre-commit hook

vpoliukh:
Added style and config files for pre-commit hook.
How to use could be seen in [Jira task #AL-10019](https://anodot.atlassian.net/browse/AL-10019)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,6 @@
+repos:
+-   repo: https://github.com/pre-commit/mirrors-yapf
+    rev: v0.30.0
+    hooks:
+    - id: yapf
+      args: ['--style=.style.yapf', '--parallel', '--in-place']
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,327 @@
+[style]
+# Align closing bracket with visual indentation.
+align_closing_bracket_with_visual_indent=True
+
+# Allow dictionary keys to exist on multiple lines. For example:
+#
+#   x = {
+#       ('this is the first element of a tuple',
+#        'this is the second element of a tuple'):
+#            value,
+#   }
+allow_multiline_dictionary_keys=False
+
+# Allow lambdas to be formatted on more than one line.
+allow_multiline_lambdas=False
+
+# Allow splitting before a default / named assignment in an argument list.
+allow_split_before_default_or_named_assigns=True
+
+# Allow splits before the dictionary value.
+allow_split_before_dict_value=False
+
+#   Let spacing indicate operator precedence. For example:
+#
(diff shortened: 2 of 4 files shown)


AntonZelenin:
I would change the following values:
coalesce_brackets=True
dedent_closing_brackets=True
split_before_arithmetic_operator=True
split_penalty_after_opening_bracket=0 or something close to 0, mb 30, not sure

Also probably would be better to keep
```
supported = [
    source.TYPE_CLICKHOUSE,
    source.TYPE_DIRECTORY,
    source.TYPE_INFLUX,
    source.TYPE_KAFKA,
    source.TYPE_MYSQL,
    source.TYPE_ORACLE,
    source.TYPE_POSTGRES
]
```
instead of
```
supported = [
    source.TYPE_CLICKHOUSE, source.TYPE_DIRECTORY, source.TYPE_INFLUX, source.TYPE_KAFKA,
    source.TYPE_MYSQL, source.TYPE_ORACLE, source.TYPE_POSTGRES
]
```
but I don't know how to achieve it

@dariakharlan what do you think?

vpoliukh:
This kind of reformat
```python
supported = [
    source.TYPE_CLICKHOUSE, source.TYPE_DIRECTORY, source.TYPE_INFLUX, source.TYPE_KAFKA,
    source.TYPE_MYSQL, source.TYPE_ORACLE, source.TYPE_POSTGRES
]
```
will happen only if initially list elements will be placed in one big single line bigger than 100 chars

AntonZelenin:
No, initially every value was on a separate line, but after reformat there were two lines like I showed. At least with my configuration. You can find `def supports_schema(pipeline_: Pipeline):` and test there

dariakharlan (changes requested):
Create a makefile command to automate the installation
Also, it would be nice to have a section in the readme file on how to set up dev environment, what commands to run (including this tool installation and rrdtool thing we had)

dariakharlan:
Yes, I agree that each item on a separate line looks better

vpoliukh:
Checked, current style saves each list item on single line if list elements end with comma, like this
```python
supported = [
    source.TYPE_CLICKHOUSE,
    source.TYPE_DIRECTORY,
    source.TYPE_INFLUX,
    source.TYPE_KAFKA,
    source.TYPE_MYSQL,
    source.TYPE_ORACLE,
    source.TYPE_POSTGRES,
]
```

Which also good, because if new element will be added to the list, then git diff will show only 1 changed line (that added element), instead of 2 lines (comma to pre-last and new added element)

AntonZelenin (approved):

dariakharlan (approved):

PR #322, opened 2021-12-07
Сreate test pipeline after installation

vpoliukh:
Added new command to the agent: 'run-test-pipeline'
How to use could be seen in [Jira task #AL-10052](https://anodot.atlassian.net/browse/AL-10052)


agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+def perform_cleanup():
+    """
+    Perform deletion of temporary source and pipeline
+    """
+    # delete temporary pipeline
+    try:
+        if e := pipeline.manager.force_delete(constants.LOCAL_RUN_TESTPIPELINE_NAME):
+            raise pipeline.PipelineException(str(e))
+    except (destination.repository.DestinationNotExists, pipeline.PipelineException):
+        pass
```
AntonZelenin, line 96: you raise PipelineException and then pass it

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+            f'Pipeline {constants.LOCAL_RUN_TESTPIPELINE_NAME} info: '
+            f'status: {info_status} '
+            f'metrics: {info_["metrics"]}'
+        )
+        if info_status != pipeline.Pipeline.STATUS_RUNNING:
+            raise pipeline.PipelineException("Error in Pipeline running. See error log for details")
+        sdc_client.stop(pipeline.repository.get_by_id(constants.LOCAL_RUN_TESTPIPELINE_NAME))
+        click.echo(f'Pipeline {constants.LOCAL_RUN_TESTPIPELINE_NAME} is stopped')
+    except (sdc_client.ApiClientException, pipeline.PipelineException) as e:
+        raise e
```
AntonZelenin, line 84: you don't need try/catch if you just reraise exceptions

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+    try:
+        if e := pipeline.manager.force_delete(constants.LOCAL_RUN_TESTPIPELINE_NAME):
+            raise pipeline.PipelineException(str(e))
+    except (destination.repository.DestinationNotExists, pipeline.PipelineException):
+        pass
+
+    # delete temporary source
+    try:
+        source.repository.delete_by_name(constants.LOCAL_RUN_TESTPIPELINE_NAME)
+    except (source.SourceException, source.repository.SourceNotExists):
```
AntonZelenin, line 101: delete_by_name doesn't raise SourceException

agent/tests/input_files/raw/pipelines/test_pipeline.json
```diff
@@ -0,0 +1,43 @@
+[
```
AntonZelenin, line 1: `input_files/raw/` is not the best place for this pipeline and its source, this folder contains configs for a specific kind of pipeline that saves data locally without transformations. Create a separate dir for them

agent/tests/test_run_test_pipeline.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_run_pipeline(cli_runner):
+    result = cli_runner.invoke(cli.run_test_pipeline, catch_exceptions=False)
+    print(f"run_test_pipeline: {result.exit_code} {result.output}")
+
+    assert result.exit_code == 0
+    assert "Source run_test_pipeline created" in result.output
+    assert "Pipeline run_test_pipeline is starting" in result.output
+    assert "Test run OK" in result.output
```
AntonZelenin, line 11: I recommend not checking for a specific text in the output, you'll need to update tests every time you change output text and if you forget you'll know about it only when the test runs on github, that takes about 10 minutes

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    try:
+        _check_prerequisites()
+        _add_source()
+        _add_pipeline()
+        _run_pipeline()
+    except Exception as e:
+        click.secho(f'Test run failed: {e}', fg='red')
+    else:
+        click.secho('Test run OK', fg='green')
```
AntonZelenin, line 27: if the pipeline got error responses from the destination it says `Test run OK` anyway
```
Pipeline run_test_pipeline info: status: RUNNING metrics: In: 20 - Out: 1 - Errors 20 (100.0%)
Test run OK
```

Also lets highlight all info messages, for example with yellow, because we show quite a lot of logs and it's hard to see important ones

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Creates temporary pipeline for run-test-pipeline command
+    """
+    pipeline_file = os.path.join(constants.LOCAL_RUN_TESTPIPELINE_DIR, "pipelines", "test_pipeline.json")
+    try:
+        with open(pipeline_file, "r") as file:
+            pipeline.json_builder.build_using_file(file)
+    except (FileNotFoundError, ValidationError, pipeline.PipelineException) as e:
+        logger.error(e)
+        raise Exception("Error during test Pipeline creation. See error log for details")
```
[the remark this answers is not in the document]
vpoliukh: Updated with ClickException

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Creates temporary source for run-test-pipeline command
+    """
+    source_file = os.path.join(constants.LOCAL_RUN_TESTPIPELINE_DIR, "sources", "test_source.json")
+    try:
+        with open(source_file, "r") as file:
+            source.json_builder.create_from_file(file)
+    except (ValidationError, SchemaError) as e:
+        logger.error(e)
+        raise Exception("Error during test Source creation. See error log for details")
```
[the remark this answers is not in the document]
vpoliukh: Updated with ClickException

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,87 @@
+import os
+import time
+import click
+import sdc_client
+
+from jsonschema import ValidationError
+from agent import destination, pipeline, source
+from agent.cli.pipeline import check_prerequisites, get_info
```
AntonZelenin, line 8: we prefer to import a module and then call its functions, like
```
from agent import cli
...
cli.pipeline.check_prerequisites()
```
it's more clear what the function does and may prevent circular dependencies in some cases

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+        raise click.ClickException("Error during temporary Pipeline creation. See error log for details")
+
+
+def _run_pipeline():
+    """
+    Runs the pipeline, get an info about delivery and stops pipeline
+    """
+    try:
+        click.echo(f'Pipeline `{constants.LOCAL_RUN_TESTPIPELINE_NAME}` is starting...')
+        pipeline_id = pipeline.repository.get_by_id(constants.LOCAL_RUN_TESTPIPELINE_NAME)
```
AntonZelenin, line 65: it should be `pipeline_`, not `pipeline_id`

agent/src/agent/cli/apply.py
```diff
@@ -28,8 +28,7 @@ def apply(work_dir, keep_not_existing):
 
 
 def populate_source_from_file(file):
-    source.check_prerequisites()
-    exceptions = []
+    exceptions = source.check_prerequisites()
```
AntonZelenin, line 31: please call `cli.source.check_prerequisites()` here, we shouldn't try to populate sources and pipelines if streamsets has not been set up and that function will raise an exception

agent/src/agent/cli/apply.py
```diff
@@ -47,18 +46,15 @@ def populate_source_from_file(file):
 
 
 def populate_pipeline_from_file(file):
-    pipeline.check_prerequisites()
-    exceptions = []
+    exceptions = pipeline.check_prerequisites()
```
AntonZelenin, line 49: same here

agent/tests/test_run_test_pipeline.py
```diff
@@ -0,0 +1,9 @@
+from agent import cli
+
+
+def test_run_pipeline(cli_runner):
+    result = cli_runner.invoke(cli.run_test_pipeline, catch_exceptions=False)
+    print(f"run_test_pipeline: {result.exit_code} {result.output}")
+
+    assert result.exit_code == 0
+    assert "Test run completed" in result.output
```
AntonZelenin, line 9: please remove this check, see a comment in run_test_pipeline()

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Creates temporary source and pipeline to check the destination is accessible.
+    """
+    try:
+        check_prerequisites()
+        _add_source()
+        _add_pipeline()
+        _run_pipeline()
+    except click.ClickException as e:
+        click.secho(f'Test run failed: {e}', fg='red')
```
AntonZelenin, line 26: if you want to make sure your command ran successfully you can call `exit(1)` here and your test will check if the exit code is 0

agent/src/agent/cli/run_test_pipeline.py
```diff
@@ -0,0 +1,103 @@
...
+def _add_source():
+    """
+    Creates temporary source for run-test-pipeline command
+    """
+    source_file = os.path.join(constants.LOCAL_RUN_TESTPIPELINE_DIR, "sources", "test_source.json")
+    try:
+        with open(source_file, "r") as file:
+            source.json_builder.create_from_file(file)
+    except source.SourceException as e:
+        _check_exist(e)
```
dariakharlan, line 45: checking text in errors can be problematic because it may change and we won't know about it. Why don't just show the error as is?

containers/streamsets/Dockerfile
```diff
@@ -15,6 +15,10 @@ RUN sudo chmod +x /tmp/configure_image.sh && /tmp/configure_image.sh
 COPY kafka-clients-2.2.2.jar ${SDC_DIST}/streamsets-libs/streamsets-datacollector-apache-kafka_2_0-lib/lib/kafka-clients-2.0.1.jar
 COPY streamsets-datacollector-kafka_multisource-protolib-3.18.0.jar  ${SDC_DIST}/streamsets-libs/streamsets-datacollector-apache-kafka_2_0-lib/lib/streamsets-datacollector-kafka_multisource-protolib-3.18.0.jar
 
+## This is needed for run-test-pipeline command
+CMD mkdir -p /home/sdc/tmp/
```
dariakharlan, line 19: First, CMD in Dockerfile is the main process that runs the container, for stuff like this RUN should be used.
Second, mkdir here shouldn't be necessary because according to the docs https://docs.docker.com/engine/reference/builder/#copy if you do COPY and if <dest> doesn’t exist, it is created along with all missing directories in its path.

dariakharlan (changes requested):

containers/streamsets/Dockerfile
vpoliukh: Fixed. Thanks

agent/src/agent/cli/run_test_pipeline.py
vpoliukh: It was just for testing. Already fixed

agent/src/agent/pipeline/pipeline.py
```diff
@@ -448,3 +446,27 @@ class PipelineRetries(Entity):
...
+
+
+class PipelineMetric(object):
+    REQUIRED_KEYS = ['counters']
+
+    def __init__(self, metrics_: dict):
+        for key in self.REQUIRED_KEYS:
+            if key not in metrics_:
+                raise KeyError(f'The key `{key}` not in input dict')
+        self.stats = {
```
AntonZelenin, line 458: there's no point in keeping these values in a dict, you can use regular attributes

agent/src/agent/pipeline/pipeline.py
```diff
@@ -448,3 +446,27 @@ class PipelineRetries(Entity):
...
+        self.stats['errors_perc'] = self.stats['errors'] * 100 / self.stats['in'] if self.stats['in'] != 0 else 0
+
+    def __str__(self):
+        return 'In: {in} - Out: {out} - Errors {errors} ({errors_perc:.1f}%)'.format(**self.stats)
+
+    def has_error(self):
+        return self.stats['errors'] > 0
+
+    def has_undelivered(self):
+        return self.stats['out'] == 0 or self.stats['out'] < self.stats['in']
```
AntonZelenin, line 472: the situation when there are fewer output metrics than input metrics is completely fine. As an example, a pipeline can filter out some input metrics, or create one Anodot metric from multiple input metrics. So in this situation you can only check if the output is empty

agent/src/agent/pipeline/pipeline.py
```diff
@@ -448,3 +446,27 @@ class PipelineRetries(Entity):
     def __init__(self, pipeline_: Pipeline):
         self.pipeline_id = pipeline_.name
         self.number_of_error_statuses = 0
+
+
+class PipelineMetric(object):
```
AntonZelenin, line 451: there's no need to inherit object, class definitions with and without object are the same

dariakharlan (approved):

PR #330, opened 2022-01-06
Working lookup for observium

AntonZelenin:


AntonZelenin:
Open questions:
1. Is `value` a good key for the function transformation?

agent/src/agent/modules/functions/transform.py
```diff
@@ -0,0 +1,39 @@
...
+    else:
+        raise Exception(f'Transformation function `{name}` is not supported')
+
+
+# todo this validation works during a pipeline run, not on creation
+def validate(func: Callable, args: list):
+    name = func.__name__
+    if name not in NUM_FUNCTION_ARGS:
+        raise Exception(f'Function `{name}` is not supported')
+    if NUM_FUNCTION_ARGS[name] != len(args):
```
dariakharlan, line 38: I think you can use inspect.signature module from a python standard library to get the number of arguments in a function [https://stackoverflow.com/questions/847936/how-can-i-find-the-number-of-arguments-of-a-python-function](url)

agent/src/agent/modules/functions/transform.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def regex_substring(string: str, pattern: str, repl: str) -> str:
+    res = re.sub(pattern, repl, string)
+    if res == 'no_match':
+        raise Exception(f'Pattern `{pattern}` didn\'t match anything for the string `{string}`, repl `{repl}`')
+    return res
+
+
+def get_by_name(name: str) -> Callable:
```
dariakharlan, line 24: you can use `getattr` to get the function from a module by it's name instead of if/else, all names match anyway. It will throw AttributeError if no such function. The same is for lookup also [https://stackoverflow.com/questions/3061/calling-a-function-of-a-module-by-using-its-name-a-string](url)

agent/src/agent/modules/functions/transform.py
```diff
@@ -0,0 +1,39 @@
...
+    if name == TO_UPPER:
+        return to_upper
+    elif name == RE_SUB:
+        return regex_substring
+    else:
+        raise Exception(f'Transformation function `{name}` is not supported')
+
+
+# todo this validation works during a pipeline run, not on creation
+def validate(func: Callable, args: list):
```
dariakharlan, line 34: this method can be used not only for transform functions, you can move it to functions module I think

agent/src/agent/data_extractor/observium/observium.py
```diff
@@ -0,0 +1,103 @@
...
+                    source_.config[source.ObserviumSource.USERNAME],
+                    source_.config[source.ObserviumSource.PASSWORD],
+                ),
+                params=params,
+                verify=source_.verify_ssl,
+                timeout=source_.query_timeout,
+            )
+            res.raise_for_status()
+            return res.json()[response_key]
+        except requests.HTTPError as e:
```
dariakharlan, line 59: we need to send source error metric to our monitoring in case of error response from observium

agent/src/agent/modules/lookup.py
```diff
@@ -0,0 +1,82 @@
...
+    res == 'Interface_name' // output: True
+    """
+    global _cache
+    if lookup_name not in _cache:
+        _cache[lookup_name] = _sources[lookup_name].get_data()
+    res = None
+    for obj in _cache[lookup_name]:
+        if compare_function(obj[key_field], lookup_value):
+            if res is not None:
+                raise Exception(
```
dariakharlan, line 57: How is it possible? Do we really need to fail all execution in this case?

dariakharlan (changes requested):

agent/src/agent/pipeline/config/stages/base.py
```diff
@@ -31,10 +30,10 @@ def _get_script(self) -> str:
             return f.read()
 
 
-class JythonScript(Stage, ABC):
+class _JythonScript(Stage, ABC):
```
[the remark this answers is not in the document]
AntonZelenin: This class is kinda internal, JythonSource and JythonProcessor inherit from it and only they should be used as there are only 2 types of Jython scripts

agent/src/agent/pipeline/pipeline.py
```diff
@@ -174,6 +174,17 @@ def dimension_names(self) -> list:
     def dimension_paths_with_names(self) -> dict:
         return dict(zip(self.dimension_paths, self.dimension_names))
 
+    @property
+    def dimension_configurations(self) -> Optional[dict]:
+        if not isinstance(self.dimensions, list):
+            raise Exception((
```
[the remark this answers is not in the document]
AntonZelenin: We can, I just didn't implement it because now it's not needed. It was quite a while ago when I last worked with optional dimensions

agent/src/agent/pipeline/config/jython_scripts/replace_illegal_chars.py
```diff
@@ -0,0 +1,50 @@
...
+            replace_all(record, 'properties')
+        if 'dimensions' in record.value:
+            replace_all(record, 'dimensions')
+        if 'measurements' in record.value:
+            replace_keys(record, 'measurements')
+        if 'tags' in record.value:
+            replace_keys(record, 'tags')
+            for k, vals in record.value['tags'].items():
+                record.value['tags'][k] = [replace_illegal_chars(v) for v in vals]
+        if 'what' in record.value:
```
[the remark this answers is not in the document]
AntonZelenin: if it's protocol 3 then it's not

agent/src/agent/modules/functions/transform.py
AntonZelenin: The thing is, a user doesn't need to provide all the arguments. For example, regex_substrig takes 3 args, but the user needs to provide only 2 because the transformed value itself is the third argument. That's why I made it as a separate configuration

agent/src/agent/modules/lookup.py
AntonZelenin: As an example it can be `startswith('port_')` and there could be two strings like `port_important` and `port_very_important`. I think in this case we cannot pick a random value from these two and there must always be only one value that matches provided criteria

agent/src/agent/pipeline/manager.py
```diff
@@ -312,3 +312,15 @@ def get_next_bucket_start(bs: str, offset: float) -> datetime:
     elif bs == pipeline.FlushBucketSize.DAY_1:
         return dt.replace(hour=0, minute=0, second=0, microsecond=0) + timedelta(days=1)
     raise Exception('Invalid bucket size provided')
+
+
+def build_dimension_configurations(dimensions: list, dimension_configurations: dict) -> dict:
```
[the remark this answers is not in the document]
AntonZelenin: indeed :)

agent/src/agent/pipeline/config/jython_scripts/replace_illegal_chars.py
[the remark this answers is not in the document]
AntonZelenin: right

agent/src/agent/pipeline/config/stages/base.py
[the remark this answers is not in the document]
dariakharlan: It's an abstract class anyway, no one will be able to use it and if you will want to inherit from it outside this module it will be weird with the underscore

agent/src/agent/modules/functions/transform.py
dariakharlan: Maybe when you call validate method you can pass all arguments? or maybe it's just always n_args-1

agent/src/agent/pipeline/pipeline.py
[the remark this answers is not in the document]
dariakharlan: ok, let's skip for now

agent/src/agent/pipeline/config/stages/base.py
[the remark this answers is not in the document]
AntonZelenin: yes, this is the plan) I don't want anyone to inherit from it outside this module

agent/src/agent/modules/lookup.py
dariakharlan: ok, but one more thing actually. Looping through every value is not efficient, maybe we also can store the result in a dict?

agent/src/agent/modules/functions/transform.py
AntonZelenin: yes, also thought about it, will use minus one

dariakharlan (approved):

PR #331, opened 2022-01-11
Fix error record creation when there is no data for a given interval

vpoliukh:
- Remove /timestamp from required fields list on javascript stage
- Added check that last_timestamp is not a single key in a record on JS stage

Related ticket: [AL-9995](https://anodot.atlassian.net/browse/AL-9995)

### DIFF
--- a/agent/src/agent/pipeline/config/js_scripts/protocol_3.js
+++ b/agent/src/agent/pipeline/config/js_scripts/protocol_3.js
@@ -64,9 +64,22 @@ function get_measurements(record) {
     return measurements
 }
 
+function last_timestamp_only(record) {
+    var record_keys = [];
+    for (var key in record) {
+        record_keys.push(key)
+    }
+    return (record_keys.length === 1 && record_keys[0] === 'last_timestamp');
+}
 
 for (var i = 0; i < records.length; i++) {
     try {
+        var timestamp_from = records[i].value['last_timestamp']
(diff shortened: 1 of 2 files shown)


AntonZelenin (approved):

agent/src/agent/pipeline/config/js_scripts/protocol_3.js
```diff
@@ -64,9 +64,22 @@ function get_measurements(record) {
...
+        record_keys.push(key)
+    }
+    return (record_keys.length === 1 && record_keys[0] === 'last_timestamp');
+}
 
 for (var i = 0; i < records.length; i++) {
     try {
+        var timestamp_from = records[i].value['last_timestamp']
+		var timestamp_to = records[i].value['last_timestamp'] + state['INTERVAL']
+        if (last_timestamp_only(records[i].value)) {
```
dariakharlan, line 79: you can move outside the loop, no need to check it for every record. Just make sure that `records.length >= 1`

dariakharlan (changes requested):

agent/src/agent/pipeline/config/js_scripts/protocol_3.js
vpoliukh: Moved outside

dariakharlan (approved):

PR #340, opened 2022-01-27
Observium migration to mysql source

AntonZelenin:


agent/src/agent/source/validator.py
```diff
@@ -250,17 +251,17 @@ class ObserviumValidator(Validator):
...
-            res = session.get(url, verify=self.source.config.get('verify_ssl', True), timeout=self.source.query_timeout)
-            res.raise_for_status()
-        except requests.exceptions.RequestException as e:
+            MySQL(
+                self.source.config['host'],
+                self.source.config.get('port', 3306),
+                self.source.config.get('username'),
+                self.source.config.get('password'),
+                self.source.config['database'],
+            ).execute('show databases')
```
dariakharlan, line 261: maybe `show tables` will be better? the user may not have permissions to show all databases

agent/src/agent/pipeline/json_builder/observium.py
```diff
@@ -120,44 +118,63 @@ class ObserviumBuilder(Builder):
...
             dim_configurations[HOST_NAME] = {field.Variable.VALUE_PATH: 'sysName'}
         if LOCATION not in dim_configurations:
             dim_configurations[LOCATION] = {field.Variable.VALUE_PATH: 'location'}
         return dim_configurations
+
+    def _validate_dimensions(self):
+        incorrect_dims = [
+            dim_name for dim_name in self.config['dimension_configurations'].keys()
+            if dim_name not in self.config['dimensions']
+        ]
```
dariakharlan, line 167: just doing `self.config['dimension_configurations'].keys() - self.config['dimensions']` will work

agent/src/agent/modules/mysql.py
```diff
@@ -0,0 +1,27 @@
...
+            user=self.user,
+            password=self.password,
+            database=self.database,
+        )
+        cursor = db.cursor()
+        cursor.execute(query)
+        columns = [i[0] for i in cursor.description]
+        data = [dict(zip(columns, row)) for row in cursor.fetchall()]
+        cursor.close()
+        db.close()
```
dariakharlan, line 26: how many queries are we doing? opening a new connection for every query might be an overhead

dariakharlan (changes requested):

agent/src/agent/source/validator.py
AntonZelenin: yep

agent/src/agent/modules/mysql.py
AntonZelenin: one query per pipeline

dariakharlan (approved):

PR #343, opened 2022-01-31
Influx multiple measures

vpoliukh:
- Added custom query in pipeline config
- Added filtering stage
- Updated Influx/Influx2 json schemas with 'query' and 'transform' fields

Related ticket: [AL-9987](https://anodot.atlassian.net/browse/AL-9987)


agent/tests/input_files/influx_pipelines.json
```diff
@@ -21,6 +21,22 @@
...
+      "required": [],
+      "optional": ["cpu", "host"]
+    },
+    "transform": {
+      "file": "/home/influx_transform.csv"
+    },
+    "properties": {"test": "val"},
+    "interval": 1200000,
+    "uses_schema": false
+  },
```
AntonZelenin, line 39: can we have a test that uses the query field?

vpoliukh: Added custom test query quite similar to auto-generated

AntonZelenin (approved):

agent/src/agent/pipeline/config/stages/source/influx.py
```diff
@@ -26,7 +26,10 @@ def get_auth_config(self):
             'conf.client.basicAuth.password': self.pipeline.source.config.get('password', '')
         }
 
-    def get_query(self):
+    def get_query(self) -> str:
+        if self.pipeline.query:
+            return f"{quote_plus(self.pipeline.query)}"
```
dariakharlan, line 31: we should add the timestamp condition so pipeline will be able to fetch them continuously. Like in jdbc pipelines, require a client to have a {TIMESTAMP_CONDITION} constant in the query and replace it with `("time" > ${record:value('/last_timestamp')} AND "time" <= ${record:value('/last_timestamp')} + {interval} AND "time" < now() - {delay})"

dariakharlan (changes requested):
`measurement_name` config in the pipeline shouldn't be required if the query is specified
`measurement_name`  here is like a table name and influx allows specifying multiple names in its query like this `select name, value from meas1, meas2, meas3 where something = 1`. Right now the agent won't parse a response from a query like that completely, we should fix it (and create a test for it)

agent/src/agent/pipeline/config/stages/source/influx.py
vpoliukh: Updated.
Maybe we need to set manually these **meas1, meas2, meas3** (from your example) values as  `measurement_name` in pipeline config so they appear in next JS stages

Because now, for example, records in JS tranform stage look as following (measurement_category - empty):

```text
properties : MAP
   target_type : gauge
   test : val
   cpu : cpu-total-usage
   host : msk-air-trn5
   measurement_category : {}
   what : usage_active
```

AntonZelenin (approved):

PR #345, opened 2022-02-02
Promql 3.0

AntonZelenin:


agent/src/agent/pipeline/json_builder/json_builder.py
```diff
@@ -127,27 +133,67 @@ def _validate_config_for_create(config: dict):
...
+    config['uses_schema'] = _uses_schema(pipeline_, config, is_edit)
     json_builder.get(pipeline_, config, is_edit).build()
     # todo too many validations, 4 validations here
     pipeline.config.validators.get_config_validator(pipeline_.source.type).validate(pipeline_)
 
 
+class _SchemaChooser:
+    @staticmethod
+    def choose(pipeline_: Pipeline, config: dict, is_edit=False) -> bool:
+        if isinstance(pipeline_, pipeline.RawPipeline):
```
dariakharlan, line 161: shouldn't we add the same condition here for TestPipeline also? if we decide that test pipelines don't use schema

agent/src/agent/pipeline/json_builder/json_builder.py
```diff
@@ -127,27 +133,67 @@ def _validate_config_for_create(config: dict):
...
+            and 'values' in config and bool(config['values'])
+        )
+        if conf_uses and not actual_schema:
+            raise ConfigurationException(
+                'PromQL pipelines support protocol 3.0 only if `dimensions` and `values` are specified'
+            )
+        return actual_schema
+
+
+def _uses_schema(pipeline_: Pipeline, config: dict, is_edit=False) -> bool:
```
dariakharlan, line 189: you should have a method that returns an instance of _SchemaChooser interface with an abstract method `choose` that you will be able to call later in the _load_config method. Right now it's not clear what a `chooser` variable can be

dariakharlan (changes requested):
use single jython source file for schema and no schema
no schema pipeline should have the same components as all other pipelines (like js transform script etc)

agent/src/agent/pipeline/json_builder/json_builder.py
AntonZelenin: it was there -_-, probably I accidentally overwrote it when force-pushed

AntonZelenin: yes, it's also like a duplicate code, I'll try to avoid it

dariakharlan (approved):

PR #354, opened 2022-02-15
[agent][observium] Allow measurement transformations

vpoliukh:
- Add watermark stage to the base pipeline
- Add measurement_configurations which allow applying transformations
- Add tag_configurations which allow use extracted values as tags

Related ticket: [AL-10467](https://anodot.atlassian.net/browse/AL-10467)


agent/src/agent/pipeline/json_builder/json_schema_definitions/observium.json
```diff
@@ -11,6 +11,7 @@
     "dimension_configurations": {"type": "object"},
     "lookup": {"type": "object"},
     "values": {"type": "object"},
+    "values_configurations": {"type": "object"},
```
AntonZelenin, line 14: `value_configurations` is correct in English, and please change it in all places

agent/src/agent/pipeline/json_builder/json_schema_definitions/observium.json
```diff
@@ -20,6 +21,7 @@
       "type": "object",
       "patternProperties":  {"[^ \\.]+": {"type": "array", "items": {"type": "string"}}}
     },
+    "tags_configurations": {"type": "object"},
```
AntonZelenin, line 24: `tag_configurations` and please change it in all places

agent/src/agent/data_extractor/observium/observium.py
```diff
@@ -31,18 +31,21 @@ def _get(pipeline_: Pipeline):
...
     schema_id = pipeline_.get_schema_id()
     try:
         for obj in data:
             metric = {
                 "timestamp": obj[pipeline_.timestamp_name],
-                "dimensions": field.extract_fields(fields, obj),
-                "measurements": _extract_measurements(obj, value_paths),
+                "dimensions": field.extract_fields(fields_dims, obj),
+                "measurements": field.extract_fields(fields_meas, obj),
+                "tags":  pipeline_.tags | {name: [tags] for name, tags in field.extract_fields(fields_tags, obj).items()},
```
AntonZelenin, line 44: `pipeline_.tags` will be added in streamsets so it's not needed here

agent/src/agent/data_extractor/observium/observium.py
```diff
@@ -31,18 +31,21 @@ def _get(pipeline_: Pipeline):
...
                 "schemaId": schema_id,
             }
+            if not metric['tags']:
```
AntonZelenin, line 47: I think it's fine if `tags` will be empty, they will be filled in streamsets

agent/src/agent/modules/functions/transform.py
```diff
@@ -16,6 +16,10 @@ def regex_substring(string: str, pattern: str, repl: str) -> str:
     return res
 
 
+def divide_by(value: int, divider: int) -> float:
```
AntonZelenin, line 19:
```suggestion
def divide(numerator: int | float, denominator: int | float) -> float:
```

agent/src/agent/modules/functions/compare.py
```diff
@@ -11,10 +12,18 @@ def equals(a, b) -> bool:
...
+def contains(needle: str, haystack: str) -> bool:
     return needle in haystack
 
 
+def regex_contains(pattern: str, s: str):
+    try:
+        pattern_ = re.compile(pattern)
+    except re.error:
+        raise Exception(f'Invalid regex pattern: {pattern}')
+    return re.search(pattern_, s)
```
AntonZelenin, line 24: shouldn't it return bool?

agent/src/agent/pipeline/json_builder/observium.py
```diff
@@ -120,6 +120,8 @@ def _load_config(self):
         self.config['dimensions'] = self._dimensions()
         self._validate_dimensions()
         self.config['values'] = self._measurements()
+        self.config['values_configurations'] = self._values_configurations()
```
AntonZelenin, line 123: could you explain why do we have value and dimension configurations here but don't have tags?

agent/src/agent/pipeline/pipeline.py
```diff
@@ -181,7 +181,15 @@ def dimension_configurations(self) -> Optional[dict]:
...
-        return _build_dimension_configurations(self.dimensions, self.config.get('dimension_configurations'))
+        return _build_transformation_configurations(self.dimensions, self.config.get('dimension_configurations'))
+
+    @property
+    def values_configurations(self) -> Optional[dict]:
+        return _build_transformation_configurations(list(self.values), self.config.get('values_configurations', {}))
+
+    @property
+    def tags_configurations(self) -> Optional[dict]:
+        return self.config.get('tags_configurations', {})
```
AntonZelenin, line 192: why don't we have build here?

agent/tests/input_files/observium/pipelines.json
```diff
@@ -3,7 +3,7 @@
...
+    "query": "select ifInBroadcastPkts, ifInErrors_rate, ifOctets_rate, ifOutUcastPkts, ifInNUcastPkts, ifUcastPkts_rate, ifInDiscards_rate, ifOutNUcastPkts_rate, ifOutBroadcastPkts_rate, ifOutMulticastPkts_rate, ifInUcastPkts, ifDiscards_rate, ifInOctets_rate, ifOutErrors, ifOutNUcastPkts, ifInOctets_perc, ifInNUcastPkts_rate, ifOutUcastPkts_rate, ifInDiscards, ifOutBroadcastPkts, ifInBroadcastPkts_rate, port_mcbc, ifInErrors, ifOutDiscards, ifOutOctets_rate, ifOutMulticastPkts, ifInUcastPkts_rate, ifOutOctets_perc, ifOutDiscards_rate, ifOutOctets, ifInMulticastPkts_rate, ifOutErrors_rate, ifInOctets, ifErrors_rate, ifInMulticastPkts, ifSpeed, ifDescr, ifAlias, port_label, poll_time, ifType, d.sysName, dl.location from ports p join devices d on d.device_id = p.device_id join devices_locations dl on dl.device_id = p.device_id",
```
AntonZelenin, line 6: could you please explain why ifDescr and ifAlias were added?

agent/src/agent/modules/field.py
```diff
@@ -73,4 +73,4 @@ def build_fields(fields_conf: dict) -> list[Field]:
 
 
 def extract_fields(fields: list[Field], data: dict) -> dict:
-    return {field_.get_name(): field_.extract_from(data) for field_ in fields}
+    return {field_.get_name(): field_.extract_from(data) for field_ in fields if field_.extract_from(data) != ''}
```
AntonZelenin, line 76: I don't think this is a good idea, first, transformations are applied two times here for every field which increases the load, second, it compares to a string, but could it be None?

agent/src/agent/data_extractor/observium/observium.py
```diff
@@ -31,18 +31,21 @@ def _get(pipeline_: Pipeline):
 def _create_metrics(data: dict, pipeline_: Pipeline) -> list:
     metrics = []
     # these values must be outside the for loop for optimization purposes
-    fields = field.build_fields(pipeline_.dimension_configurations)
-    value_paths = pipeline_.value_paths
-    value_paths = dict(zip(value_paths, value_paths))
+    fields_dims = field.build_fields(pipeline_.dimension_configurations)
+    fields_meas = field.build_fields(pipeline_.values_configurations)
```
AntonZelenin, line 35: Maybe we should call it `measurement_configurations` instead of value? Need to think about it more

agent/src/agent/data_extractor/observium/observium.py
```diff
@@ -31,18 +31,21 @@ def _get(pipeline_: Pipeline):
...
+    fields_tags = field.build_fields(pipeline_.tags_configurations)
```
AntonZelenin, line 36: could you please also generate 10 000 records of observium data in the dummy_destination and test how fast it will process them? There were some performance issues in the first release, so want to be sure it's fine. Don't need to push it, just check locally

agent/tests/input_files/observium/pipelines.json
vpoliukh: 'ifDescr' and 'ifAlias' added as tags in "tag_configurations" section

agent/src/agent/data_extractor/observium/observium.py
vpoliukh: Agree. Fixed

agent/src/agent/data_extractor/observium/observium.py
vpoliukh: Yes, you are right

agent/src/agent/pipeline/pipeline.py
vpoliukh: To allow such tags combination in a config file
```
"tags": 
    {
      "some_static_tag": ["tag_values"]
    },
    "tag_configurations": {
      "INTERFACE_ALIAS": {
        "value_path": "ifAlias"
      },
    }
```

If the build function will be used, then its output will be like this:
```
{'INTERFACE_ALIAS': {'value_path': 'ifAlias'},
 'some_static_tag': {'value_path': 'some_static_tag'}}  # and we don't have  this value_path': 'some_static_tag'
```

agent/src/agent/data_extractor/observium/observium.py
vpoliukh: Performed measurement on entire extract_metrics functiion like this:
```
    t1_start = perf_counter()
    metrics = data_extractor.observium.extract_metrics(pipeline_)
    t1_stop = perf_counter()
```
objects in data - time
150000 - 6.814 sec
30000 - 1.343 sec
9000 - 0.363 sec
900 - 0.048 sec

agent/src/agent/data_extractor/observium/observium.py
vpoliukh: Named it `value_configurations` because there is `value` in config. Maybe we should rename both? `measurements` and `measurement_configurations`?

AntonZelenin: we cannot rename both because it will break backward compatibility. I think `measurement_configurations` will be better because it describes how to create an Anodot measurement, and `value` refers only to the field in the source data that contains a value

agent/src/agent/data_extractor/observium/observium.py
AntonZelenin: nice, thanks

AntonZelenin (approved):

dariakharlan (approved):

PR #368, opened 2022-03-21
[api][pipeline] Elastic query when creating pipeline with API

vpoliukh:
- Send query with the request body when creating pipeline with API
- Update elastic source `validate_connection`

Related ticket: [AL-10746](https://anodot.atlassian.net/browse/AL-10746)


agent/src/agent/source/validator.py
```diff
@@ -102,9 +102,17 @@ class ElasticValidator(Validator):
 
     @if_validation_enabled
     def validate_connection(self):
-        # todo
-        self.source.config[source.ElasticSource.CONFIG_IS_INCREMENTAL] = False
-        super().validate_connection()
```
AntonZelenin, line 107: please use the validation that existed before

AntonZelenin (approved):

dariakharlan (changes requested):
Please also add to the api docs (agent/api/docs) an example of how to use a query parameter and  to our main wiki page

dariakharlan (approved):

PR #372, opened 2022-03-25
[agent][events] - Events agent

AntonZelenin:


dariakharlan (changes requested):
create a separate command for creating events pipeline

agent/src/agent/pipeline/config/jython_scripts/events_destination.py
```diff
@@ -0,0 +1,57 @@
...
+    import requests
+finally:
+    sdc.importUnlock()
+
+
+def is_expired(auth_token):
+    res = auth_token.split('.')[1]
+    res = base64.b64decode(res)
+    expiration_timestamp = json.loads(res)['exp']
+    return expiration_timestamp < time.time()
```
dariakharlan, line 20: i would add `expiration_timestamp < time.time() - 300` just in case

agent/src/agent/pipeline/config/jython_scripts/events_destination.py
```diff
@@ -0,0 +1,57 @@
...
+        try:
+            s = requests.Session()
+            s.headers.update({
+                'Content-Type': 'application/json',
+                'Authorization': 'Bearer ' + auth_token,
+            })
+            res = s.post(sdc.userParams['EVENTS_URL'], json=record.value, proxies=sdc.userParams['PROXIES'])
+            res.raise_for_status()
+            sdc.output.write(record)
+        except Exception as e:
```
dariakharlan, line 50: we should do retries in case of 5xx errors and timeouts

dariakharlan (changes requested):

PR #373, opened 2022-03-29
[agent][cli] Fix CLI error messages

vpoliukh:
- Add exception handling for source does not exist cases
- Add exception handling for pipeline does not exist cases
- Add the missing description for `agent pipeline ...` commands

Related ticket: [AL-10472](https://anodot.atlassian.net/browse/AL-10472)


AntonZelenin (approved):

AntonZelenin (approved):

agent/src/agent/modules/logger.py
```diff
@@ -15,6 +15,7 @@ def get_logger(name, level=logging.DEBUG, stdout=False):
 
     if stdout:
         handler = logging.StreamHandler(sys.stdout)
+        handler.setLevel(logging.ERROR)
```
dariakharlan, line 18: setting global level error is unnecessary. it means that log will show only error messages, but info messages are also fine

agent/src/agent/pipeline/json_builder/json_builder.py
```diff
@@ -39,7 +38,8 @@ def _build_multiple(configs: list, build_func: Callable) -> List[Pipeline]:
             pipeline.manager.check_pipeline_id(config['pipeline_id'])
             pipelines.append(build_func(config))
         except Exception as e:
-            exceptions[config['pipeline_id']] = f'{type(e).__name__}: {traceback.format_exc()}'
+            exceptions[config['pipeline_id']] = f'{type(e).__name__}: {e}'
+            logger_.debug(e, exc_info=True)
```
dariakharlan, line 42: this is error so it should be `logger_.error(e, exc_info=True)`. Same for others

dariakharlan (changes requested):

PR #378, opened 2022-04-05
[agent][influx] Exception when updating influx pipelines 

vpoliukh:
- Fixed Influx1/Influx2 test pipelines to process preview. For source validation it requests list of buckets, for pipeline preview requests actual data
- Updated Influx1 _edit_ CLI command for pipelines with query
- Replaced logger's exception() message by debug() for pipeline/manager. It will place just error message into console and error with trace into log file

Related ticket: [AL-11006](https://anodot.atlassian.net/browse/AL-11006)


agent/src/agent/pipeline/config/jython_scripts/test_pipelines/influx.py
```diff
@@ -39,22 +40,35 @@ def influx_request(base_url, url, method='GET', params=None, data=None, ssl=Fals
 try:
     influx_url_parsed = urlparse(sdc.userParams['HOST'])
     ssl = True if influx_url_parsed.scheme == 'https' else False
-    # get list of databases
-    params = {'q': 'SHOW DATABASES', 'db': None}
+    if sdc.userParams['MEASUREMENT_NAME']:
+        params = {'q': 'SELECT * FROM ' + sdc.userParams['MEASUREMENT_NAME'], 'db': sdc.userParams['DATABASE']}
+    elif sdc.userParams['QUERY']:
+        params = {'q': sdc.userParams['QUERY'].replace('{TIMESTAMP_CONDITION}', '1=1'), 'db': sdc.userParams['DATABASE']}
+    else:
```
dariakharlan, line 47: we should always have either QUERY or MEASUREMENT_NAME. The only case when it's not defined is when you create or edit a source. In that case, when creating a test pipeline we can set the `SHOW DATABASES` and db `null` there and get rid of these else conditions

dariakharlan, line 47: we should always have either QUERY or MEASUREMENT_NAME. The only case when it's not defined is when you create or edit a source. In that case, when creating a test pipeline we can set the `SHOW DATABASES` and db `null` there and get rid of these else conditions

agent/src/agent/cli/prompt/pipeline/influx.py
```diff
@@ -16,15 +16,25 @@ def prompt_config(self):
...
+        self.set_transform()
         self.set_uses_schema()
         self.config['timestamp'] = {
             'type': 'unix_ms',
             'name': 'time',
         }
 
+    def set_measurement_name(self):
+        if self.pipeline.query:
+            self.set_query()
```
dariakharlan, line 28: this looks weird, calling set_query in set_measurement_name. Maybe we should just delete the prompt, we don't need it anyway

dariakharlan (changes requested):

agent/src/agent/cli/prompt/pipeline/__init__.py
```diff
@@ -27,4 +26,7 @@ def get_prompter(pipeline_: Pipeline, default_config: dict, advanced: bool) -> P
         source.TYPE_SOLARWINDS: SolarWindsPrompter,
         source.TYPE_ZABBIX: ZabbixPrompter,
     }
-    return prompters[pipeline_.source.type](pipeline_, default_config, advanced)
+    try:
+        return prompters[pipeline_.source.type](pipeline_, default_config, advanced)
+    except KeyError as e:
+        raise PipelineException(f'No prompter available for Pipeline with source type `{pipeline_.source.type}`') from e
```
AntonZelenin, line 32: it should not be possible to call a source prompter that doesn't exist, they are filtered out in the _filter_without_cli, so I don't think we need to handle this exception. If a key error occurs, it's a sign for a developer something went wrong, maybe we can add a test if it doesn't exist

agent/src/agent/pipeline/config/stages/source/influx2.py
```diff
@@ -62,32 +62,22 @@ def _get_filter_condition(self) -> str:
...
-        return [
-            {
-                'key': 'URL',
-                'value': self._get_url()
-            },
-            {
-                'key': 'HEADERS',
-                'value': self._get_headers()
-            },
+        params = super()._get_script_params()
```
AntonZelenin, line 70: I don't think you should call the parent method here, query and monitoring_url from it are not used, just provide its own list of params and you also won't need to overwrite _get_query()

agent/tests/test_input/test_influx_http.py
```diff
@@ -31,35 +29,6 @@ def test_influx2_influxql_source_create(self, cli_runner):
...
-
-    def test_create_with_file(self, cli_runner, file_name, override_config):
-        super().test_create_with_file(cli_runner, file_name, override_config)
-
-    def test_edit_with_file(self, cli_runner):
-        input_file_path = get_input_file_path('influx_pipelines_edit.json')
-        result = cli_runner.invoke(cli.pipeline.edit, ['-f', input_file_path], catch_exceptions=False)
-        assert result.exit_code == 0
-
     def test_influx2_source_create(self, cli_runner):
```
AntonZelenin, line 32: please remove influx source prompter as well

AntonZelenin (changes requested):

agent/src/agent/pipeline/config/stages/source/influx2.py
vpoliukh: Updated

AntonZelenin (approved):

dariakharlan (approved):

PR #382, opened 2022-04-12
[agent] Update retry policy for RUNNING_ERROR cases

vpoliukh:
- Set pipelines retry attempts in case of RUNNING_ERROR to infinity
- Added `notification_sent` flag to `pipeline_retries` table to avoid sending notification when **send_to_bc.py** called every 1 minute and  pipeline try to start every 5 minutes
- Added separate cron script to reset `number_of_error_statuses` counter and `notification_sent` flag

Related ticket: [AL-10986](https://anodot.atlassian.net/browse/AL-10986)


agent/src/agent/pipeline/manager.py
```diff
@@ -285,7 +288,8 @@ def should_send_error_notification(pipeline_: Pipeline) -> bool:
     return not constants.DISABLE_PIPELINE_ERROR_NOTIFICATIONS \
            and pipeline_.error_notification_enabled() \
            and pipeline_.retries \
-           and pipeline_.retries.number_of_error_statuses - 1 >= constants.STREAMSETS_MAX_RETRY_ATTEMPTS
+           and pipeline_.retries.number_of_error_statuses - 1 == constants.STREAMSETS_NOTIFY_AFTER_RETRY_ATTEMPTS \
```
AntonZelenin, line 291: `>=` just in case there're more retries for some reason

agent/src/agent/pipeline/manager.py
```diff
@@ -80,12 +80,15 @@ def check_pipeline_id(pipeline_id: str):
...
     reset_pipeline_retries(pipeline_)
+    set_pipeline_retries_notification_sent(pipeline_, False)
     sdc_client.start(pipeline_, wait_for_sending_data)
 
 
 def stop(pipeline_: Pipeline):
     try:
         sdc_client.stop(pipeline_)
+        reset_pipeline_retries(pipeline_)
+        set_pipeline_retries_notification_sent(pipeline_, False)
```
AntonZelenin, line 91: I think if we reset pipeline retries, we should always reset notification as well, if so, you can put this call inside `reset_pipeline_retries`

agent/src/agent/pipeline/manager.py
```diff
@@ -331,6 +335,13 @@ def increase_retry_counter(pipeline_: Pipeline):
     pipeline.repository.save(pipeline_.retries)
 
 
+def set_pipeline_retries_notification_sent(pipeline_: Pipeline, value: bool):
+    if not pipeline_.retries:
```
AntonZelenin, line 339: I think it should not be possible to call this method when retries don't exist, so this check is probably not needed
Or at least you don't need to create retries if they don't exist

AntonZelenin (changes requested):

agent/src/agent/scripts/reset_retry_counters.py
```diff
@@ -0,0 +1,37 @@
...
+def main():
+    num_of_errors = 0
+    try:
+        pipelines = pipeline.repository.get_all()
+    except Exception:
+        _update_errors_count(0)
+        raise
+
+    for pipeline_ in pipelines:
+        try:
```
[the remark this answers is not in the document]
vpoliukh: Skipped pipeline_ with retries as None

AntonZelenin (approved):

dariakharlan (approved):

PR #383, opened 2022-04-15
[agent][monitoring] Add new metric - avg pipeline lag

vpoliukh:
- Added new metric - avg pipeline lag

Related ticket: [AL-10971](https://anodot.atlassian.net/browse/AL-10971)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ all: build-all test-all
 
 build-all: get-streamsets-libs build-docker sleep setup-all
 
-test-all: run-unit-tests test-flask-app test-streamsets test-raw-input test-raw-pipelines test-destination test-run-test-pipeline test-apply test-api test-api-scripts test-input test-export-sources test-streamsets-2 test-send-to-bc test-pipelines test-send-to-watermark
+test-all: run-unit-tests test-flask-app test-streamsets test-raw-input test-raw-pipelines test-destination test-run-test-pipeline test-apply test-api test-api-scripts test-input test-export-sources test-streamsets-2 test-send-to-bc test-pipelines test-send-to-watermark test-monitoring-metrics
 
 ##-------------
 ## DEVELOPMENT
--- a/Makefile
+++ b/Makefile
@@ -144,6 +144,9 @@ test-pipelines:
 test-run-test-pipeline:
 	$(DOCKER_TEST) tests/test_run_test_pipeline.py
 
+test-monitoring-metrics: bootstrap
+	$(DOCKER_TEST) tests/test_monitoring.py
+
 test-api:
 	$(DOCKER_TEST) tests/api/test_destination.py
 	$(DOCKER_TEST) tests/api/test_streamsets.py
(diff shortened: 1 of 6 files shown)


agent/src/agent/api/routes/pipeline.py
```diff
@@ -162,10 +163,12 @@ def pipeline_status_change(pipeline_id: str):
...
 def pipeline_offset_changed(pipeline_id: str):
-    pipeline.manager.update_pipeline_offset(
-        pipeline.repository.get_by_id(pipeline_id),
-        float(request.get_json()['offset'])
-    )
+    offset_ = float(request.get_json()['offset'])
+    pipeline_ = pipeline.repository.get_by_id(pipeline_id)
+    labels_ = (pipeline_.streamsets.url, pipeline_.name, pipeline_.source.type)
+    pipeline.manager.update_pipeline_offset(pipeline_, offset_)
+    logger.info(f'offset_: {offset_}, now: {datetime.now().timestamp()}')
```
AntonZelenin, line 170: is it very important to log this info? It might be too much data and our mechanism of enabling different log levels is not working now

agent/tests/test_monitoring.py
```diff
@@ -0,0 +1,67 @@
+import pytest
+import requests
+from .test_input.test_zpipeline_base import TestInputBase
+from .test_pipelines.test_zpipeline_base import TestPipelineBase
+
+
+class TestDirectory(TestInputBase, TestPipelineBase):
```
AntonZelenin, line 7: I guess you need to add this test to the Makefile test section, it didn't run on the github action.
And maybe we can test it in the test directory pipeline? I'm not sure it's a good idea, but that way we won't need to create directory sources and pipelines two times.
And btw, you don't delete sources and pipelines and test input files will try to create them for a second time.
And btw, it will break other directory tests because these pipelines will send data, and those pipelines will send data for the second time and test output will fail

AntonZelenin (changes requested):

agent/src/agent/api/routes/pipeline.py
vpoliukh: Sure not needed

agent/tests/test_monitoring.py
vpoliukh: Yes, firstly I thought to add into Directory tests, but then decided to create separate test (Directory sources\pipelines - because they do not require additional containers).
If in a new Makefile command we add also `bootstrap` command before execution this new test, it will not affect other tests, right?

AntonZelenin: Maybe we can create a separate single pipeline for this test?

Makefile
```diff
@@ -144,6 +144,9 @@ test-pipelines:
 test-run-test-pipeline:
 	$(DOCKER_TEST) tests/test_run_test_pipeline.py
 
+test-monitoring-metrics: bootstrap
```
AntonZelenin, line 147: add it to `test-all`

AntonZelenin (changes requested):

AntonZelenin (approved):

dariakharlan (approved):

PR #384, opened 2022-04-19
Fetch pipeline statuses on agent container start and update them in the db

AntonZelenin:


agent/src/agent/scripts/update-pipeline-statuses.py
```diff
@@ -0,0 +1,24 @@
...
+
+def main():
+    try:
+        statuses = sdc_client.get_all_pipeline_statuses()
+        for pipeline_ in pipeline.repository.get_all():
+            if pipeline_.name in statuses:
+                pipeline_.status = statuses[pipeline_.name]['status']
+                pipeline.repository.save(pipeline_)
+    except Exception:
+        monitoring.metrics.SCHEDULED_SCRIPTS_ERRORS.labels('update-pipeline-statuses.py').inc(1)
```
dariakharlan, line 19: I don't think sending monitoring metric here is the correct way to handle exceptions. If we can't save statuses to the db we should fail. If we can't get statuses from streamsets we should try several times and fail if no success

dariakharlan (changes requested):

agent/src/agent/scripts/update-pipeline-statuses.py
AntonZelenin: Well, I'm not sure it is a good idea to fail here. I added this exception handling just to be sure the agent container always can start. If there is any type of a bug/exception/behaviour change in the future that might occur in the customer environment that we didn't expect - they won't be able to start the agent container. Now the container will work, and we will still see the error in monitoring and will be able to fix it

dariakharlan: then allow an option to run without failing (turned off by default). if we have issues at this stage it's important to address it asap and sending monitoring metric can delay it or even make it unnoticeable

dariakharlan (approved):

PR #386, opened 2022-04-20
[agent][cli][apply] Remove the entity from this Session if build fails

vpoliukh:
- Remove the entity from this Session if pipeline build fails. This will allow to continue build for other pipelines 

Related ticket: [AL-11148](https://anodot.atlassian.net/browse/AL-11148)


AntonZelenin (approved):

agent/src/agent/pipeline/json_builder/json_builder.py
```diff
@@ -93,8 +97,12 @@ def edit_multiple(configs: list) -> List[Pipeline]:
...
 def edit(config: dict) -> Pipeline:
     pipeline_ = pipeline.repository.get_by_id(config['pipeline_id'])
-    _load_config(pipeline_, config, is_edit=True)
-    pipeline.manager.update(pipeline_)
+    try:
+        _load_config(pipeline_, config, is_edit=True)
+        pipeline.manager.update(pipeline_)
+    except Exception as e:
+        pipeline.repository.expunge(pipeline_)
+        raise e
```
dariakharlan, line 105: I think writing a context manager in `pipeline.repository` will be a nicer approach. It will expunge in case of exceptions or do save if all is ok. https://book.pythontips.com/en/latest/context_managers.html#implementing-a-context-manager-as-a-class

dariakharlan (changes requested):

agent/src/agent/pipeline/manager.py
```diff
@@ -123,7 +123,6 @@ def update(pipeline_: Pipeline):
     if pipeline_.uses_schema:
         _update_schema(pipeline_)
     sdc_client.update(pipeline_)
-    pipeline.repository.save(pipeline_)
```
vpoliukh, line 126: This line deleted, because `repository.save(pipeline_)` will be performed in SessionManager's context `__exit__`

PR #393, opened 2022-05-02
Directory periodic watermarks

AntonZelenin:


agent/tests/unit/test_watermark_manager.py
```diff
@@ -0,0 +1,123 @@
...
+    else:
+        pipeline_.offset = None
+        pipeline_.has_offset = Mock(return_value=False)
+
+    watermark_manager = pipeline.manager.PeriodicWatermarkManager(pipeline_)
+    watermark_manager._get_local_now_timestamp = Mock(return_value=now)
+    assert watermark_manager.should_send_watermark() == er
+
+
+def test_should_not_send_watermark():
```
dariakharlan, line 54: why is this a separate test and not a part of the test above?

dariakharlan (changes requested):

agent/src/agent/pipeline/json_builder/json_schema_definitions/kafka.json
```diff
@@ -26,7 +26,8 @@
       "type": "object",
       "properties":  {
         "bucket_size": {"type": "string", "enum": ["1m", "5m", "1h", "1d"]},
-        "delay": {"type": "integer"}
+        "delay": {"type": "integer"},
+        "timezone": {"type": "string"}
```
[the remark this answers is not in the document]
AntonZelenin: that timezone is for data, and this timezone is for watermarks, I think they might be different

agent/tests/unit/test_watermark_manager.py
AntonZelenin: maybe I should call it test_should_not_send_watermark_without_periodic_watermark_config? :)

PR #395, opened 2022-05-10
[AL-11317] Refactored SQL Server query

vpoliukh:
- Use indexed field for SQL Server query
Related ticket: [AL-11317](https://anodot.atlassian.net/browse/AL-11317)

### DIFF
--- a/agent/src/agent/pipeline/jdbc/query.py
+++ b/agent/src/agent/pipeline/jdbc/query.py
@@ -27,18 +27,21 @@ class Builder:
         return f'{query} LIMIT {pipeline.manager.MAX_SAMPLE_RECORDS}'
 
     def _get_timestamp_condition(self) -> str:
-        if query := self._get_indexed_query():
-            return query
-        return f'{self._timestamp_to_unix()} >= {self.TIMESTAMP_VALUE}' \
-               f' AND {self._timestamp_to_unix()} < {self.TIMESTAMP_VALUE} + {self.pipeline.interval}'
+        if self._supports_indexed_timestamp_condition():
+            return self._get_indexed_query()
+        return self._get_regular_query()
+
+    def _supports_indexed_timestamp_condition(self) -> bool:
+        return self.pipeline.timestamp_type in [pipeline.TimestampType.DATETIME, pipeline.TimestampType.STRING] and \
+               self.pipeline.source.type == source.TYPE_MSSQL
 
     def _get_indexed_query(self) -> Optional[str]:
-        query = None
-        if self.pipeline.timestamp_type in [pipeline.TimestampType.DATETIME, pipeline.TimestampType.STRING]:
-            if self.pipeline.source.type == source.TYPE_MSSQL:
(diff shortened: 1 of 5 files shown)


agent/src/agent/pipeline/jdbc/query.py
```diff
@@ -27,18 +27,21 @@ def _get_preview_query(self):
...
-               f' AND {self._timestamp_to_unix()} < {self.TIMESTAMP_VALUE} + {self.pipeline.interval}'
+        if self._supports_indexed_timestamp_condition():
+            return self._get_indexed_query()
+        return self._get_regular_query()
+
+    def _supports_indexed_timestamp_condition(self) -> bool:
+        return self.pipeline.timestamp_type in [pipeline.TimestampType.DATETIME, pipeline.TimestampType.STRING] and \
+               self.pipeline.source.type == source.TYPE_MSSQL
 
     def _get_indexed_query(self) -> Optional[str]:
```
AntonZelenin, line 38: It's not optional :)

AntonZelenin (approved):

agent/src/agent/pipeline/jdbc/query.py
vpoliukh: Sure)

agent/src/agent/pipeline/validators/jdbc_query.py
```diff
@@ -1,8 +1,12 @@
-from agent import pipeline
+from agent.pipeline.jdbc.query import TIMESTAMP_CONDITION, TIMESTAMP_COLUMN, LAST_TIMESTAMP_VALUE, INTERVAL
 
 
 def get_errors(query: str):
     errors = []
-    if pipeline.jdbc.query.TIMESTAMP_CONDITION not in query:
-        errors.append(f'Please add {pipeline.jdbc.query.TIMESTAMP_CONDITION} constant to the query')
+    if TIMESTAMP_CONDITION not in query:
+        errors.append(f'Please add {TIMESTAMP_CONDITION} constant to the query')
```
AntonZelenin, line 7: if there's no TIMESTAMP_CONDITION but other fields are present, the validation will fail, but as I understand it should fail only if there are no fields at all

AntonZelenin (changes requested):

agent/src/agent/pipeline/validators/jdbc_query.py
vpoliukh: There is errors.pop() in else statement for inner 'if'
So, if there are no TIMESTAMP_CONDITION and no [TIMESTAMP_COLUMN, LAST_TIMESTAMP_VALUE, INTERVAL] -> we will have two errors.
But if no TIMESTAMP_CONDITION and else present 'NO-TIMESTAMP' error will be popped

AntonZelenin: just don't put the error there if you don't need it

AntonZelenin: I mean change the code so that errors are added only in cases where there certainly is an error

PR #397, opened 2022-05-20
[agent][metrics] Send watermark_delta and watermark_sent metrics in all 3.0 pipelines

vpoliukh:
- Add new stage: SendMonitoringMetrics
- Update base pipelines with the new stage

Jira ticket: [AL-10894](https://anodot.atlassian.net/browse/AL-10894)


agent/src/agent/pipeline/config/jython_scripts/send_watermark_and_monitoring_metrics.py
```diff
@@ -0,0 +1,53 @@
...
+
+    sys.path.append(os.path.join(os.environ['SDC_DIST'], 'python-libs'))
+    import requests
+    import pytz
+finally:
+    sdc.importUnlock()
+
+
+def _get_watermark_delta(watermark):
+    tz = sdc.userParams['TIMEZONE'] if sdc.userParams['WATERMARK_IN_LOCAL_TIMEZONE'] == 'True' and sdc.userParams['TIMEZONE'] else 'UTC'
```
AntonZelenin, line 20: if the timezone is not specified for the pipeline it will be UTC by default

AntonZelenin (approved):

agent/src/agent/pipeline/config/jython_scripts/send_watermark_and_monitoring_metrics.py
```diff
@@ -0,0 +1,53 @@
...
+    return str(calendar.timegm(datetime.now(pytz.timezone(tz)).timetuple()) - watermark)
+
+
+for record in sdc.records:
+    if not record.value['watermark'] or not record.value['schemaId']:
+        sdc.log.info('The pipeline doesn\'t have watermark or schemaId, skipping sending watermark')
+        continue
+
+    try:
+        # send Watermark
```
dariakharlan, line 30: Let's add a flag to pipeline `watermark_logs_enabled` (false by default) and if true add here before sending the watermark something like `sdc.log.info('Sending watermark: '   str(record.value))`. It will be useful for debugging pipelines and especially if there are different timezone settings

dariakharlan (changes requested):

agent/src/agent/pipeline/config/jython_scripts/send_watermark_and_monitoring_metrics.py
```diff
@@ -38,11 +41,16 @@ def _get_watermark_delta(watermark):
             raise requests.RequestException(err_msg)
 
         # send Watermark Metrics
+        watermark_delta = _get_watermark_delta(record.value['watermark'])
+        if sdc.userParams['WATERMARK_LOGS'] == 'True':
```
dariakharlan, line 45: no need for logging metrics sending, just the watermark itself :)

dariakharlan (changes requested):

dariakharlan (approved):

PR #400, opened 2022-05-24
Delete trailing slashes in destination url

AlexMuliar:
https://anodot.atlassian.net/browse/AL-10210

Delete trailing slashes in destination manager


AntonZelenin (approved):

AntonZelenin:
And please add a unit test that will check that destinations return their URL without trailing slashes

dariakharlan (approved):

PR #403, opened 2022-05-30
[agent][streamsets] Move credential store files to SDC_DATA

vpoliukh:
- `/data/security` is a new path for jks files
- Dockerfile updated

Related ticket: [AL-11436](https://anodot.atlassian.net/browse/AL-11436)

### DIFF
--- a/containers/streamsets/credential-stores.properties
+++ b/containers/streamsets/credential-stores.properties
@@ -49,7 +49,7 @@ credentialStore.jks.config.keystore.type=PKCS12
 
 # The location of the Java keystore. Specify an absolute path or a path relative to the
 # $SDC_CONF directory.
-credentialStore.jks.config.keystore.file=jks-credentialStore.pkcs12
+credentialStore.jks.config.keystore.file=/data/security/jks-credentialStore.pkcs12
 
 # The password to access the Java keystore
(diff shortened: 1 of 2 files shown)


AntonZelenin (approved):

containers/streamsets/credential-stores.properties
```diff
@@ -49,7 +49,7 @@ credentialStore.jks.config.keystore.type=PKCS12
 
 # The location of the Java keystore. Specify an absolute path or a path relative to the
 # $SDC_CONF directory.
-credentialStore.jks.config.keystore.file=jks-credentialStore.pkcs12
+credentialStore.jks.config.keystore.file=/data/security/jks-credentialStore.pkcs12
```
dariakharlan, line 52: there won't be any errors if /data/security directory doesn't exist?

vpoliukh: Will be an error. That is what I've added one more step with mkdir command to a Dockerfile

vpoliukh: I've also tried to specify env variable instead of `/data/security/`, but it did not work. You can see attached screenshot in a Jira ticket

dariakharlan (approved):

PR #408, opened 2022-06-07
AL-11277: Filter metrics for old streamsets

AlexMuliar:


agent/src/agent/monitoring/__init__.py
```diff
@@ -18,15 +19,18 @@ def pull_latest():
...
     for metric in metrics.registry.collect():
         target_type = anodot.TargetType.COUNTER if metric.type == 'counter' else anodot.TargetType.GAUGE
         for sample in metric.samples:
             if sample.name.endswith('_created'):
                 continue
-            if 'pipeline_id' in sample.labels and sample.labels['pipeline_id'] not in pipeline_ids:
+            if 'pipeline_id' in sample.labels and sample.labels['pipeline_id'] not in pipelines:
+                continue
+            if streamsets_url := sample.labels.get('streamsets_url') and\
+                                 streamsets_url != pipelines[sample.labels['pipeline_id']].streamsets.url:
```
dariakharlan, line 32: `sample.labels['pipeline_id']` might not exist so it can fail with KeyError. Please check if it exists before the if condition

PR #410, opened 2022-06-08
AL-11286: set level from env var or default

AlexMuliar:
Tiket [AL-11286](https://anodot.atlassian.net/browse/AL-11499)


agent/src/agent/modules/logger.py
```diff
@@ -5,8 +5,23 @@
...
+    'DEBUG': logging.DEBUG,
+    'NOTSET': logging.NOTSET,
+}
+
+
+def get_logger(name, level=None, stdout=False):
     logger = logging.getLogger(name)
+
+    if not level:
+        level = levels.get(os.environ.get('LOGGER_LEVEL'), logging.DEBUG)
```
dariakharlan, line 24: you can use `logging.getLevelName` function instead of defining your own dict 
and plz change the default level to INFO
And the constant name should be `LOG_LEVEL`, I think it's a more common name for things like that

dariakharlan (changes requested):

agent/src/agent/modules/logger.py
vpoliukh: I think you don't need to duplicate levels and, for example, declaring a function like this should work fine
`def get_logger(name, level=os.environ.get("LOG_LEVEL", logging.INFO), stdout=False):`

agent/src/agent/modules/logger.py
```diff
@@ -5,8 +5,21 @@
...
+    'WARNING': logging.WARNING,
+    'INFO': logging.INFO,
+    'DEBUG': logging.DEBUG,
+    'NOTSET': logging.NOTSET,
+}
+
+
+def get_logger(name, level=None, stdout=False):
     logger = logging.getLogger(name)
+    level = level or levels.get(os.environ.get('LOG_LEVEL'), logging.INFO)
```
vpoliukh, line 22: Levels are duplicated with default logging lib. How about this:
`level = level or os.environ.get('LOG_LEVEL', logging.INFO)`

Next function `logger.setLevel(level)` accepts both string and int values

AlexMuliar: What about typoff case?

vpoliukh: For example, LOG_LEVEL == 'critical'. 
In this case of suggested change, we will have an exception, like: `Unknown level: 'critical'`
In case of double get() we will have INFO level, which is confusing I think

AlexMuliar: Ok, I will use try-except for this case

vpoliukh (approved):
I think you can set `LOG_LEVEL: 'INFO'` for docker-compose.yml and `LOG_LEVEL: 'DEBUG'` for docker-compose-dev.yml

dariakharlan (approved):

PR #411, opened 2022-06-15
[agent][monitoring]AL-11398: get metrics async

AlexMuliar:


agent/src/agent/monitoring/streamsets.py
```diff
@@ -68,8 +71,42 @@ def _pull_kafka_metrics(streamsets_: streamsets.StreamSets):
...
+    if not asynchronous:
+        for pipeline_ in pipelines:
+            _pull_pipeline_metrics(pipeline_)
+    else:
+
+        jmxes = sdc_client.get_jmxes_async([
+            (pipeline_.streamsets, f'metrics:name=sdc.pipeline.{pipeline_.name}.*',)
+            for pipeline_ in pipelines])
+        for pipeline_, jmx in zip(pipelines, jmxes):
+            _pull_pipeline_metrics(pipeline_, jmx)
```
vpoliukh, line 96: As I right understand, that only jmx changes depending on asynchronous flag. So, how about preparing jmxes first and then running pulling.
Like so:

```python
def _process_pipeline_metrics(pipelines: List[pipeline.Pipeline], asynchronous: bool = False) -> None:
    # prepeare jmxes
    if not asynchronous:
        jmxes = [sdc_client.get_jmx(pipeline_.streamsets, f'metrics:name=sdc.pipeline.{pipeline_.name}.*')
                 for pipeline_ in pipelines]
    else:
        jmxes = sdc_client.get_jmxes_async([(pipeline_.streamsets, f'metrics:name=sdc.pipeline.{pipeline_.name}.*',)
                                            for pipeline_ in pipelines])
    # actual pulling
    for pipeline_, jmx in zip(pipelines, jmxes):
        _pull_pipeline_metrics(pipeline_, jmx)
```

This will make to request for jmxes in on function instead of two different

agent/src/agent/monitoring/streamsets.py
```diff
@@ -57,8 +57,8 @@ def _pull_pipeline_metrics(pipeline_: pipeline.Pipeline):
                 bean['999thPercentile'] / 1000)
 
 
-def _pull_kafka_metrics(streamsets_: streamsets.StreamSets):
-    jmx = sdc_client.get_jmx(
+def _pull_kafka_metrics(streamsets_: streamsets.StreamSets, jmx=None):
+    jmx = jmx or sdc_client.get_jmx(
```
vpoliukh, line 61: Same as above. jmx could be None here?

agent/src/agent/monitoring/streamsets.py
```diff
@@ -28,8 +29,7 @@ def _is_influx(pipeline_: pipeline.Pipeline):
     return pipeline_.source.type == source.TYPE_INFLUX
 
 
-def _pull_pipeline_metrics(pipeline_: pipeline.Pipeline):
-    jmx = sdc_client.get_jmx(pipeline_.streamsets, f'metrics:name=sdc.pipeline.{pipeline_.name}.*')
+def _pull_pipeline_metrics(pipeline_: pipeline.Pipeline, jmx=None):
```
vpoliukh, line 32: I guess jmx is not an optional parameter now?
for bean in jmx['beans'] will raise TypeError in case of None.

vpoliukh (approved):

dariakharlan (approved):

PR #415, opened 2022-06-20
AL-11533: get_number_of_arguments by comma or regex

AlexMuliar:


dariakharlan (changes requested):
please add here more explicit test use cases
1. Add unit test for the regex that will check different inputs and expected results
2. add values with comas that need to be transformed to the e2e tests (to the source data and to the expected output)

dariakharlan (approved):

PR #423, opened 2022-07-08
AL-10979 no data notification

AlexMuliar:


agent/src/agent/pipeline/manager.py
```diff
@@ -124,6 +127,13 @@ def reset_pipeline_retries(pipeline_: Pipeline):
         pipeline.repository.save(pipeline_.retries)
 
 
+def reset_pipeline_notifications(pipeline_: Pipeline):
+    if pipeline_.notifications:
+        if pipeline_.notifications.no_data_notification:
```
vpoliukh, line 132: Why not a single line if?
`if pipeline_.notifications and pipeline_.notifications.no_data_notification:`

vpoliukh (commented):
Could we add some tests to see how it work?
As I remember there are some pipeline which have no data, just watermarks (test_sage_schema_file_dvp)

agent/alembic/versions/971cd06e3d76_create_pipline_notifications_table.py
```diff
@@ -0,0 +1,27 @@
+"""create pipline notifications table
```
dariakharlan, line 1: what is the purpose of this table? why we can't join directly with no_data_notifications?

agent/src/agent/pipeline/manager.py
```diff
@@ -313,14 +323,30 @@ def transform_for_bc(pipeline_: Pipeline) -> dict:
...
 
 
+def should_send_no_data_error_notification(pipeline_: Pipeline) -> bool:
+    if pipeline_.notifications and pipeline_.notifications.no_data_notification:
+        no_data_notification_period = timedelta(
+            minutes=pipeline_.notifications.no_data_notification.notification_period
+        ) + timedelta(seconds=pipeline_.interval or 0)
+        return should_send_error_notification(pipeline_) \
+               and not pipeline_.notifications.no_data_notification.notification_sent \
+               and bool(pipeline_.watermark) \
```
dariakharlan, line 345: this should be compared by last sent offset (sometimes we sent watermark even if there is no data, for example, when dvp is enabled)

dariakharlan (changes requested):

agent/alembic/versions/971cd06e3d76_create_pipline_notifications_table.py
AlexMuliar: If we will extend pipeline notifications, it will be aggregated in notification attribute

dariakharlan (approved):

PR #425, opened 2022-07-14
[agent][async] Feature pipeline upd async

vpoliukh:
Added asynchronous option for CLI commands:
- `streamsets balance`
- `pipeline update`
- `apply`

Related ticket: [AL-10018](https://anodot.atlassian.net/browse/AL-10018)


agent/tests/test_pipelines/test_2/test_zabbix_http.py
```diff
@@ -27,3 +28,17 @@ def test_stop(self, cli_runner, name=None, check_output_file_name=None):
 
     def test_output_schema(self, name=None, pipeline_type=None, output=None):
         pytest.skip()
+
+    def test_start(self, cli_runner, name, sleep):
+        super().test_start(cli_runner, name, sleep)
+
+    def test_output(self, name, pipeline_type, output):
+        def compare_output():
```
dariakharlan, line 36: I think you can use `super().test_output` here instead of duplicating the code

dariakharlan (changes requested):

agent/tests/test_pipelines/test_2/test_zabbix_http.py
vpoliukh: Agree. Fixed

dariakharlan (approved):

PR #426, opened 2022-07-20
[agent][impala] Impala integration

vpoliukh:
- Added as new JDBC handler

Related ticket: [AL-11373](https://anodot.atlassian.net/browse/AL-11373)


agent/src/agent/pipeline/config/handlers/factory.py
```diff
@@ -22,6 +22,7 @@ def _get_no_schema_handler(pipeline_: Pipeline, base_config: dict) -> NoSchemaCo
     handlers_protocol20 = {
         source.TYPE_CACTI: pipeline.config.handlers.cacti.CactiConfigHandler,
         source.TYPE_ELASTIC: pipeline.config.handlers.elastic.ElasticConfigHandler,
+        source.TYPE_IMPALA: pipeline.config.handlers.jdbc.JDBCConfigHandler,
```
dariakharlan, line 25: please add also to the _get_raw_handler
and you can also add there MSSQL

dariakharlan (changes requested):

agent/src/agent/pipeline/config/handlers/factory.py
vpoliukh: Added

dariakharlan (approved):

PR #431, opened 2022-07-26
[agent][logs] AL-11775: Fix agent log rotation

AlexMuliar:


agent/src/agent/modules/logger.py
```diff
@@ -10,7 +10,9 @@ def get_logger(name, level=None, stdout=False):
     level = level or os.environ.get('LOG_LEVEL', logging.INFO)
     logger.setLevel(level)
 
-    file_handler = RotatingFileHandler(os.environ.get('LOG_FILE_PATH', 'agent.log'), maxBytes=2000)
+    file_handler = RotatingFileHandler(os.environ.get('LOG_FILE_PATH', 'agent.log'),
+                                       maxBytes=int(os.environ.get('LOG_MAX_BYTES', 2000)),
```
AlexMuliar, line 14: Maybe will be better set this variable in Kb, and name it as MAX_LOG_KBYTES or something like this

vpoliukh (approved):
Approved. But I suggest to increase file size to smth like (25mb) close to mail limit. And also set backupCount to smth like 5-10 to preserve more logs

PR #439, opened 2022-08-05
AL-11839: Add preferred type to streamsets

AlexMuliar:


agent/tests/test_streamsets_with_preferred_type.py
```diff
@@ -0,0 +1,42 @@
+import time
```
vpoliukh, line 1: `time` not used

agent/tests/test_streamsets_with_preferred_type.py
```diff
@@ -0,0 +1,42 @@
...
+    }
+    result = cli_runner.invoke(cli.streamsets.edit, [url], catch_exceptions=False, input=generate_input(input_))
+    streamsets.repository.get_by_url(url)
+    assert result.exit_code == 0
+
+
+def test_run_pipeline(cli_runner):
+    result = cli_runner.invoke(cli.run_test_pipeline, args=['--no-delete'], catch_exceptions=False)
+    print(f"run_test_pipeline: {result.exit_code} {result.output}")
+    assert pipeline.repository.get_by_streamsets_url(url)
```
vpoliukh, line 33: What we are asserting here? List not empty?

AlexMuliar: At this case, db contain 2 streamsets, and we check that pipeline with directory type is assigned to streamsets with directory type

vpoliukh: I mean usually, with `assert` statements we evaluate some expression. Smth like:
```python
assert len(pipeline.repository.get_by_streamsets_url(url)) == 1
```

AlexMuliar: Ok, I have idea how rework this test

agent/src/agent/cli/streamsets.py
```diff
@@ -34,11 +34,12 @@ def list_():
 @click.option('--username', type=click.STRING, default='admin')
 @click.option('--password', type=click.STRING, default='admin')
 @click.option('--agent-ext-url', type=click.STRING, default='http://anodot-agent')
-def add(url, username, password, agent_ext_url):
+@click.option('--preferred-type', type=click.STRING, default=None)
+def add(url, username, password, agent_ext_url, preferred_type):
```
dariakharlan, line 38: please also add prompt option here, for configuring streamsetsts we use cli mostly

dariakharlan (changes requested):

agent/tests/test_streamsets_with_preferred_type.py
```diff
@@ -0,0 +1,77 @@
...
+        with pytest.raises(streamsets.repository.StreamsetsNotExistsException):
+            streamsets.repository.get_by_url(url)
+        assert result.exit_code == 0
+
+
+def perform_cleanup():
+    try:
+        pipeline.manager.force_delete(constants.TEST_RUN_PIPELINE_NAME)
+    except (destination.repository.DestinationNotExists, pipeline.PipelineException):
+        pass
```
vpoliukh, line 69: This could be rewritten as
```python
with contextlib.suppress(destination.repository.DestinationNotExists, pipeline.PipelineException):
    pipeline.manager.force_delete(constants.TEST_RUN_PIPELINE_NAME)
```

agent/src/agent/cli/streamsets.py
```diff
@@ -166,6 +170,18 @@ def _prompt_agent_external_url(streamsets_: StreamSets) -> str:
     return url
 
 
+@infinite_retry
+def _prompt_preferred_type(streamsets_: StreamSets) -> str:
+    type_ = click.prompt('Enter streamsets preferred type', type=click.STRING)
+    validate_preferred_type(type_)
+    return type_
```
vpoliukh, line 177: Why do we pass `streamsets_` to function if it is not used?

vpoliukh (approved):

dariakharlan (approved):

PR #459, opened 2022-09-19
AL-12024: Create jdbc event pipeline

AlexMuliar:


agent/src/agent/pipeline/config/loader.py
```diff
@@ -101,6 +101,12 @@ def _check_pipeline(self, pipeline_: Pipeline):
     def _get_config_file(cls, pipeline: Pipeline) -> str:
         return {
             source.TYPE_DIRECTORY: 'directory.json',
+            source.TYPE_CLICKHOUSE: 'jdbc.json',
+            source.TYPE_DATABRICKS: 'jdbc.json',
+            source.TYPE_IMPALA: 'jdbc.json',
+            source.TYPE_MSSQL: 'jdbc.json',
+            source.TYPE_MYSQL: 'jdbc.json',
+            source.TYPE_POSTGRES: 'jdbc.json',
```
dariakharlan, line 109: Missing TYPE_ORACLE

dariakharlan (changes requested):

agent/tests/test_pipelines/test_2/test_mysql_http.py
```diff
@@ -43,13 +45,15 @@ class TestMySQL(TestPipelineBase):
             {'name': 'test_watermark_local_timezone', 'output': 'jdbc_timezone.json', 'pipeline_type': 'mysql'},
             {'name': 'test_mysql_advanced', 'output': 'jdbc_file_full.json', 'pipeline_type': 'mysql'},
             {'name': 'test_jdbc_mysql_tag', 'output': 'jdbc_mysql_tags.json', 'pipeline_type': 'mysql'},
+            # {'name': 'test_events_jdbc', 'output': 'events_jdbc.json', 'pipeline_type': 'mysql'},
```
vpoliukh, line 48: So, we do not check the expected output?

agent/tests/test_input/test_2/test_mysql_http.py
```diff
@@ -39,6 +43,9 @@ def test_create_advanced(self, cli_runner, name, source):
                                    input=f'{source}\n{name}\nSELECT * FROM test WHERE {{TIMESTAMP_CONDITION}} AND not country is NULL\n\n86400\n{days_to_backfill}\n1\ntimestamp_unix\nunix\ny\ntest\nclicks:gauge impressions:gauge\nadsize country\nkey1:val1 key2:val2\n\n\n\n\n')
         assert result.exit_code == 0
 
+    def test_create_events_pipeline_with_file(self, cli_runner, file_name, override_config: dict):
+        super(TestMySQL, self)._test_create_with_file(cli_runner, file_name, override_config, cli.pipeline.create_events)
```
vpoliukh, line 47: Why do you prefer `super(TestMySQL, self)._test_create_with_file`?
Because there is no overriding, you can call just `self._test_create_with_file()`

agent/src/agent/pipeline/manager.py
```diff
@@ -181,7 +181,7 @@ def update_source_pipelines(source_: Source):
 def update_pipeline_offset(pipeline_: Pipeline, timestamp: float):
     offset = sdc_client.get_pipeline_offset(pipeline_)
     if not offset:
-        return
+        offset = '{}'
```
vpoliukh, line 184: Could you please explain why we need this instead of a `return` before?

agent/src/agent/pipeline/config/jython_scripts/events_destination.py
```diff
@@ -17,9 +17,18 @@
 def main():
     for record in sdc.records:
         send_event(record)
+        send_offset(record.value['offset'])
```
vpoliukh, line 20: Just to be sure, if inside `send_event` function `client.send_events` will fail to send with some exception, we still perform `send_offset` after it, right?

agent/src/agent/pipeline/config/jython_scripts/events_destination.py
```diff
@@ -20,6 +20,14 @@ def main():
         sdc.output.write(record)
 
 
+def send_offset(offset):
+    try:
+        res = requests.post(sdc.userParams['AGENT_OFFSET_URL'], json={'offset': offset})
+        res.raise_for_status()
+    except (requests.ConnectionError, requests.HTTPError) as e:
+        sdc.error.write(str(e))
```
vpoliukh, line 28: If I'm not mistaken, `sdc.error` takes two arguments: record and some message.
If you want just log the error, the `sdc.log.error` should be used
[Here](https://docs.streamsets.com/datacollector/latest/help/datacollector/UserGuide/Processors/Jython.html?hl=sdc.error) is a docs

vpoliukh (approved):

dariakharlan (approved):

PR #460, opened 2022-09-21
AL-12056: Send metrics directly

AlexMuliar:


vpoliukh (approved):

agent/src/agent/monitoring/__init__.py
```diff
@@ -20,23 +20,19 @@ def get_monitoring_metrics() -> list[dict]:
...
-                    pipelines.get(sample.labels.get('pipeline_id')) and\
-                    sample.labels.get('streamsets_url') != pipelines[sample.labels['pipeline_id']].streamsets.url:
-                continue
-            dims = {**sample.labels, 'host_name': constants.HOSTNAME, 'source': 'agent_monitoring'}
-            data.append(
-                anodot.Metric20(sample.name, sample.value, target_type, datetime.utcnow(), dimensions=dims).to_dict()
-            )
-
+    # pipelines: Dict[str, Pipeline] = dict(zip({p.name for p in pipelines}, pipelines))
+    for metric in streamsets.get_metrics():
```
dariakharlan, line 24: in the registry we have not only streamnsets metrics, there also pipeline metrics that are being sent with API. So you should have  both -t he conversion of jmx for streamsets metrics and also taking metrics from the registry as it was before

dariakharlan (changes requested):

dariakharlan (approved):

PR #461, opened 2022-09-22
[jdbc][databrix] Backfill missing intervals option

vpoliukh:
Added `track_missed_intervals` bool option for JDBC pipelines

This works as follows:
- If `track_missed_intervals == True`, last offset requested from DB
- If `db_offset` is behind current offset for at least two intervals -> firstly try to look at these intervals in the past. Request data starting from this `db_offset`.

Related ticket: [AL-12057](https://anodot.atlassian.net/browse/AL-12057)

### DIFF
--- a/agent/src/agent/api/routes/pipeline.py
+++ b/agent/src/agent/api/routes/pipeline.py
@@ -162,15 +162,18 @@ def pipeline_status_change(pipeline_id: str):
     return jsonify('')
 
 
-@pipelines.route('/pipeline-offset/<pipeline_id>', methods=['POST'])
+@pipelines.route('/pipeline-offset/<pipeline_id>', methods=['POST', 'GET'])
 @needs_pipeline
 def pipeline_offset_changed(pipeline_id: str):
-    offset_ = float(request.get_json()['offset'])
     pipeline_ = pipeline.repository.get_by_id(pipeline_id)
-    labels_ = (pipeline_.streamsets.url, pipeline_.name, pipeline_.source.type)
-    pipeline.manager.update_pipeline_offset(pipeline_, offset_)
-    monitoring.metrics.PIPELINE_AVG_LAG.labels(*labels_).set(datetime.now().timestamp() - offset_)
-    return jsonify('')
+    if request.method == 'POST':
+        offset_ = float(request.get_json()['offset'])
+        labels_ = (pipeline_.streamsets.url, pipeline_.name, pipeline_.source.type)
+        pipeline.manager.update_pipeline_offset(pipeline_, offset_)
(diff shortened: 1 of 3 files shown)


agent/src/agent/pipeline/config/jython_scripts/jdbc.py
```diff
@@ -30,6 +31,36 @@ def to_timestamp(date):
...
 
+def check_missing_intervals(cur_offset, interval):
+    db_offset = get_db_offset()
+    if not db_offset:
+        return
+    while ((cur_offset - db_offset) / interval.total_seconds()) > 1:
+        sdc.log.debug(
+            'PROCESSING MISSED DATA: '
+            'from ' + str(db_offset) + ' to ' + str(db_offset + interval.total_seconds())
+        )
```
dariakharlan, line 42: in a situation where there is missing data for one day but data for the next day is present the pipeline will be stuck because the query interval is always 1 day. So in this case `date from` should be the db offset but `date to` should be the end of the current interval.

agent/src/agent/pipeline/config/jython_scripts/jdbc.py
```diff
@@ -52,6 +83,10 @@ def main():
             time.sleep(2)
             if sdc.isStopped():
                 return
+
+        if sdc.userParams['TRACK_MISSED_INTERVALS'] == 'True':
```
dariakharlan, line 87: in the case of a 1-day interval missing data will be only checked once a day, but a more efficient way would be querying missing data every hour in this case, because the data may just arrive a bit later than usual. We can add a `query_missing_data_interval` config for example

dariakharlan (changes requested):

agent/src/agent/api/routes/pipeline.py
```diff
@@ -162,15 +162,18 @@ def pipeline_status_change(pipeline_id: str):
...
-    pipeline.manager.update_pipeline_offset(pipeline_, offset_)
-    monitoring.metrics.PIPELINE_AVG_LAG.labels(*labels_).set(datetime.now().timestamp() - offset_)
-    return jsonify('')
+    if request.method == 'POST':
+        offset_ = float(request.get_json()['offset'])
+        labels_ = (pipeline_.streamsets.url, pipeline_.name, pipeline_.source.type)
+        pipeline.manager.update_pipeline_offset(pipeline_, offset_)
+        monitoring.metrics.PIPELINE_AVG_LAG.labels(*labels_).set(datetime.now().timestamp() - offset_)
+        return jsonify('')
+    elif request.method == 'GET':
```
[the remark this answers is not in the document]
vpoliukh: Fixed

dariakharlan (approved):

PR #469, opened 2022-10-27
[agent][snmp] Fix historical offset usage

vpoliukh:
- Use only the current offset value
- Separate snmp_query for SNMP Get command and query_timeout for request timeout to data_extractor

### DIFF
--- a/agent/src/agent/pipeline/config/jython_scripts/snmp.py
+++ b/agent/src/agent/pipeline/config/jython_scripts/snmp.py
@@ -26,15 +26,10 @@ def get_now():
 
 
 def main():
-    if sdc.lastOffsets.containsKey(entityName):
-        offset = int(float(sdc.lastOffsets.get(entityName)))
-    else:
-        offset = get_now()
-
+    offset = get_now()
     sdc.log.info('OFFSET: ' + str(offset))
 
     while True:
-        now = get_now()
         if sdc.isStopped():
             return
         while offset > get_now():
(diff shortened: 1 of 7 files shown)


agent/src/agent/pipeline/config/jython_scripts/snmp.py
```diff
@@ -45,7 +40,7 @@ def main():
                 return
 
         batch = sdc.createBatch()
-        res = requests.get(sdc.userParams['SNMP_SOURCE_URL'], timeout=60)
+        res = requests.get(sdc.userParams['SNMP_SOURCE_URL'], timeout=45)
```
dariakharlan, line 43: shouldn't this timeout be the value from query_timeout parameter in the source?

dariakharlan (changes requested):

agent/src/agent/pipeline/config/jython_scripts/snmp.py
vpoliukh: There are two places where timeouts used:

1. This requests.get
2. and the other is used during real SNMP command execution: 
https://github.com/anodot/daria/blob/fix-snmp-offset/agent/src/agent/data_extractor/snmp/snmp.py#L138

**query_timeout** parameter from the source used in the second

You think we should have timeouts as query_timeout parameter in both request?

dariakharlan: I think we can create a separate parameter for this timeout. It should be configurable in case when a user has many hosts to query and it takes a lot of time to process the data

dariakharlan (approved):

PR #470, opened 2022-10-28
AL-12131: Select available docker compose in system

AlexMuliar:


scripts/agent.sh
```diff
@@ -10,16 +9,27 @@ function check_cmd() {
...
+  check_cmd "docker-compose --version"
+  COMPOSE=(docker-compose)
+fi
+echo "Using \"$COMPOSE\" for script"
+
+set -e
 
 if [[ $1 == 'install' ]]; then
-  docker-compose pull && docker-compose up -d
+  $COMPOSE pull && docker compose up -d
```
dariakharlan, line 28: `docker compose up -d` should be replaced with $COMPOSE

scripts/agent.sh
```diff
@@ -10,16 +9,27 @@ function check_cmd() {
...
 elif [[ $1 == 'upgrade' ]]; then
-  docker-compose pull && docker-compose up -d && sleep 30 && docker exec -i anodot-agent agent pipeline update
+  $COMPOSE pull && docker compose up -d && sleep 30 && docker exec -i anodot-agent agent pipeline update
```
dariakharlan, line 30: `docker compose up -d` should be replaced with $COMPOSE up -d

dariakharlan (changes requested):

PR #471, opened 2022-10-28
AL-12144: fix event jdbc pipeline no data

AlexMuliar:


agent/src/agent/pipeline/config/jython_scripts/create_events.py
```diff
@@ -6,13 +6,27 @@
 finally:
     sdc.importUnlock()
 
+
+def validate_event(event, record):
+    keys = [key for key in event]
+    if len(keys) == 0:
+        return False
+    for required_key in sdc.userParams.get("REQUIRED_FIELDS", []):
```
dariakharlan, line 14: and what happens when no data? the record will not be validated because when no data there is only last timespamp

dariakharlan (changes requested):

agent/src/agent/pipeline/config/jython_scripts/create_events.py
AlexMuliar: This method validates only events, if event have no filelds, than record contain only last timestamp, and we just skips this record

agent/src/agent/pipeline/config/jython_scripts/create_events.py
```diff
@@ -6,13 +6,27 @@
 finally:
     sdc.importUnlock()
 
+
+def validate_event(event, record):
+    keys = [key for key in event]
```
vpoliukh, line 11: This could be simplified to just `keys = list(event)`

agent/src/agent/pipeline/config/jython_scripts/create_events.py
```diff
@@ -6,13 +6,27 @@
...
+    if len(keys) == 0:
+        return False
+    for required_key in sdc.userParams.get("REQUIRED_FIELDS", []):
+        if required_key not in keys:
+            sdc.error.write(record, "Record Error: {} is required".format(sdc.userParams["REQUIRED_FIELDS"]))
```
vpoliukh, line 16: Maybe it worth show in error which exactly required_key is missing instead of all REQUIRED_FIELDS?

vpoliukh (commented):
- Add minor comments.
- In general, this validate_event applied only for JDBC events and why not for Directory events? We will have the same problem if some data in event-directory-pipeline will be without required fields

vpoliukh (approved):

PR #472, opened 2022-10-31
[agent][cli] Refactored `restore` command

vpoliukh:
- Separated `restore` to `restore database` and `restore streamsets` commands
- Add exception handling for `pipeline list` for cases when some streamsets not available

### DIFF
--- a/agent/src/agent/cli/backup.py
+++ b/agent/src/agent/cli/backup.py
@@ -38,21 +38,25 @@ def database(dump_file):
 
 
 @restore.command()
-def streamsets():
+@click.option('--use-available', is_flag=True, default=False)
+def streamsets(use_available: bool):
     """
     Restore StreamSets states with its pipelines from DataBase.
     """
-    _restore_pipelines()
+    _restore_pipelines(use_available)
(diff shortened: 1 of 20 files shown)


agent/src/agent/cli/backup.py
```diff
@@ -10,24 +10,49 @@
...
+    click.secho('StreamSets pipelines successfully restored', fg='green')
 
 
 def _restore_pipelines():
     existing, not_existing = _get_pipelines()
     for pipeline_ in not_existing:
         click.echo(f'Creating pipeline `{pipeline_.name}`')
+        if not sdc_client.get_pipeline_streamsets_stat(pipeline_):
+            click.secho(f'StreamSets `{pipeline_.streamsets.get_url()}` not available', err=True, fg='red')
+            continue
```
dariakharlan, line 55: let's add an option to put the pipeline to any available streamsets in case the one it was assigned to is not available

dariakharlan (changes requested):

dariakharlan (approved):

PR #475, opened 2022-11-03
[agent] AL-12097: create or update transform file on agent apply

AlexMuliar:


agent/src/agent/cli/apply.py
```diff
@@ -51,6 +51,9 @@ def populate_pipeline_from_file(file):
         try:
             if 'pipeline_id' not in config:
                 raise Exception(f'Pipeline configs must contain a `pipeline_id` field, check {file.name}')
+            if config.get('transform'):
+                with open(config['transform']['file'], 'w+') as f:
+                    f.write(config['transform']['config'])
```
vpoliukh, line 56: Need to check that 'file' key in config['transform'], will be an exception if transform declared without file.
Sees snmp_table pipeline for example

vpoliukh (changes requested):

dariakharlan:
In the ticket I wrote ```export csv config with pipelines to a file (configs like transform files etc)``` it's not implemented here. I meant to export the transform file (for some pipelines it can be also a query file) with config when we do `agent pipeline export`. Let me know if I need to explain it more

dariakharlan (changes requested):
In the ticket I wrote ```export csv config with pipelines to a file (configs like transform files etc)```, it means exporting the transform file (for some pipelines it can be also a query file) with config when we do `agent pipeline export`. It's not implemented here, let me know if I need to explain more

agent/src/agent/cli/pipeline.py
```diff
@@ -339,14 +340,31 @@ def export(dir_path):
     """
     Export pipelines' config into file
     """
+
+    def _export_file(path, pipeline_name):
```
dariakharlan, line 344: the file might not exist anymore inside the container (for example if the container is recreated) you need to copy the content of the file from the config that is stored in the dataabse

agent/src/agent/cli/pipeline.py
```diff
@@ -339,14 +340,31 @@ def export(dir_path):
...
+        destination = os.path.join(os.path.dirname(dir_path), 'processing_files', pipeline_name)
```
dariakharlan, line 345: the file name should be the same as original value, destination path is the same dir

dariakharlan (changes requested):

dariakharlan (approved):

PR #477, opened 2022-11-04
[agent][promql] AL-11340: - validate url and the query before creating source/pipeline

AlexMuliar:


agent/src/agent/pipeline/json_builder/json_builder.py
```diff
@@ -55,6 +55,7 @@ def _build_multiple(configs: list, build_func: Callable) -> List[Pipeline]:
 
 def build(config: dict) -> Pipeline:
     _validate_config_for_create(config)
+    source_ = source.repository.get_by_name(config.get('source'))
```
vpoliukh, line 58: Seems `source_` is not used

vpoliukh (approved):
Approve with small comment

agent/src/agent/source/validator.py
```diff
@@ -260,6 +260,20 @@ def validate_token(self):
 class PromQLValidator(Validator):
     VALIDATION_SCHEMA_FILE = 'promql.json'
 
+    def validate_connection(self):
+        try:
+            url = urllib.parse.urljoin(
+                self.source.config['url'],
+                'api/v1/labels'
+            )
+            response = requests.get(url)
```
dariakharlan, line 269: add timeout to the request. If host is not available it will just hang

agent/src/agent/pipeline/config/validators/validators.py
```diff
@@ -31,6 +34,23 @@ def validate(pipeline_):
...
 
+class PromQLValidator(Validator):
+    @staticmethod
+    def validate(pipeline_):
+        try:
+            query = pipeline_.config['query']
+            url = pipeline_.source_.config["url"] + '/api/v1/query?' + urllib.parse.urlencode({
+                'query': query.encode('utf-8'),
+            })
+            response = requests.get(url)
```
dariakharlan, line 45: add timeout to the request. If host is not available it will just hang

dariakharlan (changes requested):

PR #483, opened 2022-12-19
AL-12219: use specific pipeline on use elastic schema

AlexMuliar:


agent/src/agent/pipeline/config/stages/source/elastic.py
```diff
@@ -0,0 +1,59 @@
...
+            {
+                'key': "QUERY",
+                'value': self._load_query_from_file()
+            }
+        ]
+
+
+# class TestElasticScript(JythonSource):
+#     JYTHON_SCRIPT = 'elastic_loader.py'
+#     JYTHON_SCRIPTS_DIR = os.path.join(JythonSource.JYTHON_SCRIPTS_DIR, 'test_pipelines')
```
vpoliukh, line 59: I suppose we do not need this?

agent/src/agent/pipeline/config/stages/source/elastic.py
```diff
@@ -0,0 +1,59 @@
...
+                'value': str(self.pipeline.delay)
+            },
+            {
+                'key': 'QUERY_TIMEOUT',
+                'value': str(self.pipeline.source.query_timeout)
+            },
+            {
+                # Do we need to customize it?
+                'key': 'SCROLL_TIMEOUT',
+                'value': '5m'
```
vpoliukh, line 44: As I right understand, the `scroll` parameter (passed to the search request and to every scroll request) tells Elasticsearch how long it should keep the search context alive. Do we really need this?
To receive a bigger amount of hits, we could pass `"size": 10000` into the request payload, for example

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,117 @@
...
+                batch.process(entityName, str(end))
+            end += interval
+        except Exception:
+            sdc.log.error(traceback.format_exc())
+            raise
+
+
+def get_base_url():
+    return sdc.userParams['URLs'][0] + '/' + sdc.userParams['INDEX'] + "//_search?scroll=" \
+           + sdc.userParams['SCROLL_TIMEOUT']
```
vpoliukh, line 110: How about using `urljoin` here?
And if `source.config['conf.httpUris']` has multiple host we still are making requests only to first one

dariakharlan, line 110: if you use scroll parameter here you need to paginate over the results  https://www.elastic.co/guide/en/elasticsearch/reference/current/paginate-search-results.html#scroll-search-results

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,117 @@
...
+            sdc.log.debug(curr_url)
+            res = make_request(curr_url, get_query_params(end)).json()
+            sdc.log.debug(str(res))
+            if res['hits']['total']['value']:
+                process_batch(res, end)
+            elif sdc.userParams['DVP_ENABLED'] == 'True':
+                # records with last_timestamp only
+                batch = sdc.createBatch()
+                record = sdc.createRecord('record created ' + str(datetime.now()))
+                record.value = dict(res['hits']['hits'])
```
dariakharlan, line 99: what is this? we should have last_timestamp here to send the watermark

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,117 @@
...
+            continue
+        break
+    return res
+
+
+def process_batch(result_, end_):
+    batch = sdc.createBatch()
+    for res in result_['hits']['hits']:
+        # res['sort'] = [res['_source']['timestamp_unix_ms']]
+        del res['_score']
```
dariakharlan, line 66: why are you deleting this? not all search results might return this and in case this field doesn't exist there's gonna be an exception

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
dariakharlan: you can use python client that can help you handle this logic [https://github.com/elastic/elasticsearch-py](https://github.com/elastic/elasticsearch-py)

dariakharlan: [connecting to multiple nodes example](https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/connecting.html#connect-url)

agent/src/agent/pipeline/config/stages/source/elastic.py
```diff
@@ -0,0 +1,59 @@
...
+from agent.pipeline.config.stages.base import JythonSource
+
+
+class ElasticScript(JythonSource):
+    JYTHON_SCRIPT = 'elastic_loader.py'
+
+    def _load_query_from_file(self):
+        with open(self.pipeline.config['query_file']) as f:
+            query = f.read()
+            return query.replace('${OFFSET}', '"OFFSET"')
```
dariakharlan, line 13: now we are not required to use ${OFFSET} so we can ask the client to use something json compatible like `$OFFSET` and you won't need to replace it two times

agent/src/agent/pipeline/config/stages/source/elastic.py
dariakharlan: we should keep the scroll logic but it should handle it in the script, we don't want to limit requests by size and the response can be quite big and memory consuming

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,117 @@
...
+def make_request(url_, params={}):
+    for i in range(1, N_REQUESTS_TRIES + 1):
+        try:
+            sdc.log.debug(url_)
+            res = requests.post(
+                url_,
+                json=params,
+                stream=True,
+                headers={'Content-Type': 'application/json'},
+                # timeout=sdc.userParams['QUERY_TIMEOUT']
```
dariakharlan, line 48: we need timeouts always otherwise if the host is not available it will just hang forever

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
dariakharlan: and yes, you need to add sending watermark logic also

dariakharlan (changes requested):

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
dariakharlan: test dvp enabled, no data from elastic - watermark still sent

dariakharlan:
use JSConvertMetrics30 stage

agent/src/agent/pipeline/config/handlers/elastic.py
```diff
@@ -15,6 +15,18 @@ class ElasticConfigHandler(NoSchemaConfigHandler):
     }
 
 
+class ElasticSchemaConfigHandler(SchemaConfigHandler):
+    stages_to_override = {
+        'source': stages.source.elastic.ElasticScript,
+        'JavaScriptEvaluator_01': stages.js_convert_metrics.JSConvertMetrics,
```
dariakharlan: JSConvertMetrics30 is for schema

dariakharlan: you need to write tests also

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,116 @@
...
+                batch.process(entityName, str(end))
+            end += interval
+        except Exception:
+            sdc.log.error(traceback.format_exc())
+            raise
+
+
+def get_client():
+    return elasticsearch.Elasticsearch(
+        sdc.userParams['URLs'][0]
```
dariakharlan, line 108: use all urls, why only one?

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,116 @@
...
+
+def get_interval():
+    return int(sdc.userParams['INTERVAL'])
+
+
+def make_request(client, index, params={}, scroll=None):
+    for i in range(1, N_REQUESTS_TRIES + 1):
+        try:
+            sdc.log.debug(index)
+            res = client.search(body=params, index=index)
```
dariakharlan, line 44: you need to use scroll function to be able to retrieve large search results

agent/tests/input_files/elastic_query_schema.json
```diff
@@ -0,0 +1,4 @@
+{
+  "sort": [{"timestamp_unix_ms": {"order": "asc"}}],
+  "query": {"range": {"timestamp_unix": {"gt": $OFFSET}}}
```
dariakharlan, line 3: lets make it in quotes so json will be valid
```
"query": {"range": {"timestamp_unix": {"gt": "$OFFSET"}}}
```

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
dariakharlan: don't forget to specify timeouts

dariakharlan (changes requested):

agent/src/agent/pipeline/config/jython_scripts/elastic_loader.py
```diff
@@ -0,0 +1,127 @@
...
+                    process_batch(batch, end)
+                elif sdc.userParams['DVP_ENABLED'] == 'True':
+                    # records with last_timestamp only
+                    batch = sdc.createBatch()
+                    record = sdc.createRecord('record created ' + str(datetime.now()))
+                    record.value = dict()
+                    record.value['last_timestamp'] = end - get_interval()
+                    batch.add(record)
+                    batch.process(entityName, str(end))
+                end += interval
```
dariakharlan, line 111: this shouldn't be inside the loop. in the loop we process data from the same interval

dariakharlan (changes requested):

dariakharlan (approved):

PR #486, opened 2023-01-11
AL-12296: fix elastic pipeline export

AlexMuliar:


agent/src/agent/cli/pipeline.py
```diff
@@ -373,9 +372,13 @@ def _export_config_to_file(filename, config, pipeline_name):
                 pipeline_name=pipeline_.name,
             )
         if pipeline_config.get('query_file'):
+            if pipeline_.source.type == TYPE_ELASTIC:
+                query_file_config = pipeline_config['override_source'][ElasticSource.CONFIG_QUERY]
```
dariakharlan, line 376: in the new elastic version we won't have this., the query is taken from file as usual

dariakharlan (changes requested):

agent/src/agent/cli/pipeline.py
dariakharlan: please also add a change to the query validator, do not use offset field from the source, use the timestamp to check conditions
here: agent.pipeline.validators.elastic_query.get_errors

dariakharlan (approved):

PR #489, opened 2023-03-09
PIL-4279: fix dependency, broken elastic user and pass

AlexMuliar:


agent/src/agent/source/source.py
```diff
@@ -279,14 +279,23 @@ class ElasticSource(APISource):
...
 
     def set_config(self, config):
         super().set_config(config)
+        self._update_legacy_fields()
         if 'query_interval_sec' in self.config:
             self.config[ElasticSource.CONFIG_QUERY_INTERVAL] = \
                 '${' + str(self.config['query_interval_sec']) + ' * SECONDS}'
         self.config[ElasticSource.CONFIG_IS_INCREMENTAL] = True
 
+    def _update_legacy_fields(self) -> dict:
```
AntonZelenin, line 293: it doesn't return dict

agent/src/agent/source/source.py
```diff
@@ -279,14 +279,23 @@ class ElasticSource(APISource):
...
+        if ElasticSource.CONFIG_USERNAME not in self.config and APISource.USERNAME in self.config:
```
AntonZelenin, line 294: leave a comment why we're doing it

AntonZelenin (approved):

